PR #85, opened 2015-01-05
Use angle vector duration when no time is set.

aginika:
Copy the function from jsk_calibration and use it.
Tested on baxter

PR #86, opened 2015-01-12
(pr2eus/robot-interface.l): Add validity check for tm in angle-vector

orikuma:
When using limb-controller, a user have to define tm by oneself because of optional argument order in current robot-interface.
If a user send false input like
```
(send *ri* :angle-vector <float-vector> <limb-controller>)
```
, max joint velocity is stored in tm, angle-vector is sent to fullbody-controller and this causes unintended whole-body motion in maximum joint velocity.
To prevent uninteded motion, check validity of tm and if it is incorrect, angle-vector would be better to return without any effects.

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -160,6 +160,11 @@
      msg))
   (:angle-vector
    (av &optional (tm 3000) (ctype controller-type) (start-time 0))
+   ;; check validity of tm
+   (unless (numberp tm)
+     (warn ";; interpolation time is not a number: ~A~%" tm)
+     (return-from :angle-vector)
+     )
    ;; check max-joint-velocity
    (let ((idx 0)
          (diff-av (v- av (or (send self :state :potentio-vector) (send robot :angle-vector)))))
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -162,7 +162,7 @@
    (av &optional (tm 3000) (ctype controller-type) (start-time 0))
    ;; check validity of tm
    (unless (numberp tm)
-     (warn ";; interpolation time is not a number: ~A~%" tm)
+     (ros::ros-error "interpolation time is not a number: ~A" tm)
      (return-from :angle-vector)
      )
    ;; check max-joint-velocity


garaemon:
`tm` has default value, does it really work?

YoheiKakiuchi:
実際にどんなコマンドを打って、どう動いたのかな？

orikuma:
I want to prevent tm from being stored max-joint-velocity unintendedly.
I checked following situation using hrpsys-simulator. What other situation do I have to check?
``` send *ri* :angle-vector (send *staro* :reset-pose) ```
-> move fullbody in default tm
``` send *ri* :angle-vector (send *staro* :reset-pose) 1000 ```
-> move fullbody in 1000msec
``` send *ri* :angle-vector (send *staro* :reset-pose) 1000 :larm-controller ```
-> move only larm in 1000msec
``` send *ri* :angle-vector (send *staro* :reset-pose) :larm-controller ```
-> error (;; interpolation time is not a number: :larm-controller)

**Comment by YoheiKakiuchi** [comment]
Is this bug of function mux ?
~~~
8.irteusgl$ max :fuga 10.0
10.0
9.irteusgl$ max 10.0 :fuga
irteusgl 0 error: number expected in (max 10.0 :fuga)
~~~

**Comment by garaemon** [comment]
@YoheiKakiuch good catch!
I think so

2015年1月12日月曜日、Yohei Kakiuchi<notifications@github.com>さんは書きました:

> Is this bug of function mux ?
>
> 8.irteusgl$ max :fuga 10.0
> 10.0
> 9.irteusgl$ max 10.0 :fuga
> irteusgl 0 error: number expected in (max 10.0 :fuga)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/86#issuecomment-69530457>.
>


-- 
from iPhone

File: pr2eus/robot-interface.l [hunk 1]
```diff
@@ -160,6 +160,11 @@
      msg))
   (:angle-vector
    (av &optional (tm 3000) (ctype controller-type) (start-time 0))
+   ;; check validity of tm
+   (unless (numberp tm)
+     (warn ";; interpolation time is not a number: ~A~%" tm)
```
k-okada: better to use ROS_ERROR so that we can catch them via rosconsole

garaemon: In addition to that, I prefer to use error.
we should stop program here

k-okada: yes, that would be better.

--
◉ Kei Okada

garaemon:
What does tm=nil mean?
1. just a bug
2. User hopes to move robot with max speed

snozawa:
I could not fully understand these discussion, but 
I think we need to use ``error`` instead of ``ros::ros-error``.

garaemon:
> I think we need to use error instead of ros::ros-error.

I think `ros::ros-error` + `error` is best

garaemon:
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/85
ちょうどこれとコンフリクトする話題ですね。

aginika:
If we consider about #85 too,
we can assume below 4 types of input of angle-vector.
- Normal time designated(check the time is too fast or not)

 example: `( send *ri* :angle-vector (send *robot* :angle-vector) 1000)`

- Optimized time using 

 example: `(send *ri* :angle-vector (send *robot* :angle-vector) :optimize)`

- No disignated time | Safe time using (use the time 5 times of optimized speed)

 example: `(send *ri* :angle-vector (send *robot* :angle-vector))`
- Error arguments passed

 example: `( send *ri* :angle-vector (send *robot* :angle-vector) :larm-controller)`
 => error and ros::ros-error 


If we use above process,
we will do like below.
We set the tm value nill as default.

- check `tm` is nil or not
- check `tm` is number or :optimize
- check `tm` is slower than optimized time

I will change my Pull Request as follow above process.

garaemon:
@aginika 
LGTM. and I think it does not effect existing code so much.

k-okada:
what is `optimized`? is it `fastest` or `reasonable`? is it dangerous to
use `optimized` for nil input?

--
◉ Kei Okada

aginika:
It is fastest.
I think some robots will be dangerous to use optimized for some mistake.
so I set the safety.
Is is useless to prepare safe mode?

k-okada:
Ok,,, so do you have any idea than using :optimize keyword? -1, or 0???more confusing than :optimize?

aginika:
Ok, anything will do. I will use 0 for using optimized time.

snozawa:
> more confusing than :optimize?

I think 0 or -1 is more confusing than ``:optimize`` or ``nil``.

@agnika
why do you need both optimize mode or safe mode?
If optimize is dangerous, we rarely use optimize mode.

aginika:
In PR2 and Baxter, any speed is not dangerous and we don't care about arm speed.
When we use Baxter, if we give very fast time, Baxter will give up to follow the trajectory.
But we want Baxter to move their arm as fast as he can.
So we need the optimized time which will be fast and possible to execute for baxter.

I don't know whether the optimized time will not be dangerous or not for other robot.
If other robot wil not have any problem, I can delete the safe mode.

garaemon:
I prefer :optimize to -1


2015年1月14日水曜日、Yuto Inagaki<notifications@github.com>さんは書きました:

> In PR2 and Baxter, any speed is not dangerous and we don't care about arm
> speed.
> When we use Baxter, if we give very fast time, Baxter will give up to
> follow the trajectory.
> But we want Baxter to move their arm as fast as he can.
> So we need the optimized time which will be fast and possible to execute
> for baxter.
>
> I don't know whether the optimized time will not be dangerous or not for
> other robot.
> If other robot wil not have any problem, I can delete the safe mode.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/86#issuecomment-69876829>.
>


-- 
from iPhone

aginika:
If nobody has suggestion about this any more, I will update PR(#85) to change to use :optimize instread of 0. Is that OK?

garaemon:
I prefer :optimize to 0, because the following code may cause unexpected behavior:

```lisp
(send *ri* :angle-vector av (/ 10000 10001))
```

k-okada:
ok, so
`tm (time to goal in msec, if not specified, use 5 [sec]. 1/3 of the fastest speed for :optimized and fastest speed for :fastest`
please add method documentation just after `(:angle-vector` defninition

garaemon:
solved via https://github.com/jsk-ros-pkg/jsk_pr2eus/issues/85

snozawa:
@furushchev  and @aginika 

Sorry for very late discussion,
but I'd like to return to original discussion:
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/86#issuecomment-69718776

I'd like to make deprecate this argument:
```
      ((null tm)
       (setq tm (* 5 fastest-tm)))
```

, because safe or not is quite robot-dependent and `(* 5 fastest-tm)` is not always safe.

For example, assume that someone writes codse in which a robot with slow joint speed (`*pepper*`?) moves hand `#F(100 0 0)` with `tm nil` argument 
and applies the same codes to `*staro*` or `*jaxon*` (faster robots). 
Even if the same codes are used  with `tm nil` argument, 
the hand cartesian speed for `*staro*` or `*jaxon*` becomes quite faster. 
By using the same arguments and same codes, 
`tm nil` argument is safe for `*pepper*`, but dangerous for `*staro*` or `*jaxon*`. 

We also need to consider both joint-space speed limit and cartesian-space speed limit for safety.


For safety, if users set `tm` as `nil`, it would be better to invoke euslisp error. 
```
(cond
  ;;Fastest time Mode                                                                                                                                                                                                                                                                                                  
  ((equal tm :fast)
    (setq tm fastest-tm))
  ;;Normal Number disgnated Mode                                                                                                                                                                                                                                                                                       
  ((numberp tm)
    (if (< tm fastest-tm)
    (setq tm fastest-tm)))
  ;;Error Not Good Input                                                                                                                                                                                                                                                                                               
  (t
    (ros::ros-error ":angle-vector tm is invalid. args: ~A" tm)
    (error ":angle-vector tm is invalid. args: ~A" tm)))
```

According to https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/86#issuecomment-69718776 discussion, 
my proposal is here:

- Normal time designated(check the time is too fast or not)

 example: `( send *ri* :angle-vector (send *robot* :angle-vector) 1000)`

- Optimized time using 

 example: `(send *ri* :angle-vector (send *robot* :angle-vector) :fast)`
 Users can configure fastness margin using `scale` argument:
 example: `(send *ri* :angle-vector (send *robot* :angle-vector) :fast :scale 10)`

- Error arguments passed

 example: `( send *ri* :angle-vector (send *robot* :angle-vector) :larm-controller)`
 example: `( send *ri* :angle-vector (send *robot* :angle-vector) nil)`
 example: `( send *ri* :angle-vector (send *robot* :angle-vector))`
 => error and ros::ros-error

PR #90, opened 2015-01-25
Remove packages from .rosinstall on travis

garaemon:


garaemon:
```
FAILURE: Test node [roseus/roseus] does not exist or is not executable
```

garaemon:
roseus executable is not installed???

k-okada:
this is overwritten by #91

k-okada:
humm that is strange, roseus is installed, so it is strange....

garaemon:
need to re-source bashrc?

✉︎ Ryohei

PR #97, opened 2015-01-28
[pr2eus] Add make-robot-interface-from-name function to create robot-interface instance from name

garaemon:
see https://github.com/jsk-ros-pkg/jsk_visualization/issues/227

It's important to write code independent from robot!

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -892,4 +892,15 @@
           (nreverse tms))
     ))
 
+(defun make-robot-interface-from-name (name &rest args)
+  "make a robot model from string: (make-robot-model \"pr2\")"
+  (let ((name (string-downcase name)))
+    (let ((klasses (flatten (send robot-interface :hierarchy))))
+      (dolist (klass klasses)
+        (let ((klass-name (string-downcase (send klass :name))))
+        (if (or (string= klass-name name)
+                (string= (format nil "~A-interface" name) klass-name))
+            (return-from make-robot-interface-from-name (instance* klass :init args)))))
+      nil)))
+
 (provide :robot-interface "robot-interface.l")
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -894,13 +894,16 @@
 
 (defun make-robot-interface-from-name (name &rest args)
   "make a robot model from string: (make-robot-model \"pr2\")"
-  (let ((name (string-downcase name)))
-    (let ((klasses (flatten (send robot-interface :hierarchy))))
-      (dolist (klass klasses)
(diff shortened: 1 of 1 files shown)


snozawa:
To create each robot interface, we tipically write the following codes:
```
(defclass hrp2jsk-interface
  :super rtm-ros-robot-interface
  :slots ())
(defmethod hrp2jsk-interface
  (:init (&rest args)
         (send-super* :init :robot hrp2jsk-robot args)))

(defun hrp2jsk-init (&rest args)
  (if (not (boundp '*ri*))
      (setq *ri* (instance* hrp2jsk-interface :init args)))
  (if (not (boundp '*hrp2jsk*))
      (setq *hrp2jsk* (instance hrp2jsk-robot :init))))
```
But, I think this code can be replaced like this:
```
(defun hrp2jsk-init (&rest args)
  (if (not (boundp '*ri*))
      (setq *ri* (instance* robot-interface :init :robot hrp2jsk-robot args)))
  (if (not (boundp '*hrp2jsk*))
      (setq *hrp2jsk* (instance hrp2jsk-robot :init))))
```
In this case, we do not need to define each robot interface inheriting robot-interface class.

garaemon:
You mean we do not need subclass of robot-interface?

snozawa:
Sorry, robot dependent interface subclass is necessary...
So, please neglect the above comment:https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/97#issuecomment-71825671

pr2eus/robot-interface.l
```diff
@@ -892,4 +892,15 @@
           (nreverse tms))
     ))
 
+(defun make-robot-interface-from-name (name &rest args)
+  "make a robot model from string: (make-robot-model \"pr2\")"
+  (let ((name (string-downcase name)))
+    (let ((klasses (flatten (send robot-interface :hierarchy))))
+      (dolist (klass klasses)
+        (let ((klass-name (string-downcase (send klass :name))))
+        (if (or (string= klass-name name)
```
snozawa: I'd like to keep searching rules simple.
Why isn't it sufficient to search [robot-name]-interface?
How about to define function like this?
```
(let ((klass (read-from-string (format nil "~A-interface" robot-name))))
  (if (and (boundp klass) (subclassp (eval klass) robot-interface))
      (instance* (eval klass) :init args)
    (error "No such subclass of robot-interface : ~A~%" klass)))
```

garaemon: looks good!

garaemon:
updated

k-okada:
why https://github.com/euslisp/jskeus/blob/master/irteus/irtutil.l#L378   is not used in here?
can we remove `make-robot-model-from-name` ?

PR #98, opened 2015-01-29
modify :angle-vector-sequence to use :angle-vector-duration

aginika:
From https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/85#issuecomment-71589199

modiied and now we can use as below
```
(send *ri* :angle-vector-sequence (list av1 av2 av3) (list :fast 4000 nil))

(send *ri* :angle-vector-sequence (list av1 av2 av3) (list :fast :fast :test))
[ERROR] [1422533494.154349018]: :angle-vector-sequence tm is invalid.  args: 1000.0
irteusgl 0 error:  :angle-vector-sequence tm is invalid. args: 1000.0 in (error ":angle-vector-sequence tm is invalid. args: ~A" tm)
```

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -191,17 +191,16 @@
                      msg))
      msg))
 
-  (:angle-vector-duration 
-   (end &optional (scale 5) (min-time 1.0))
+  (:angle-vector-duration
+   (start end &optional (scale 5) (min-time 1.0))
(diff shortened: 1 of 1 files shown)


garaemon:
To support following usage would be great:
```
(send *ri* :angle-vector-sequence (list av1 av2 av3) :fast)
```

PR #100, opened 2015-02-09
- need to call roseus at the end of find_package so that roseus.cmake can read all package files

k-okada:


garaemon:
oops , sorry I merged colliding pull request.
Please merge origin/master

k-okada:
sound_play is already set in -> https://github.com/k-okada/jsk_pr2eus/blob/fix_compile/pr2eus/catkin.cmake#L13

garaemon:
I see, I understand finally that location of roseus is the most important.

garaemon:
After that, could you release newer version?

PR #104, opened 2015-02-15
[pr2eus] modify robot-interface.l to support control_msgs::SingleJointPo...

aginika:
...sitionGoal

For baxter head_action_server, we need to support singleJointPositionGoal


k-okada:
please fix to baxter servers, I prefer to merge this PR when we found at least two robot that uses this interface.

aginika:
>>please fix to baxter servers,
Sorry, I couldn't understand what you mean....
I think you said we need to support the SimpleJointPositionAction, is it right?
Should I close this PR?

PR #106, opened 2015-02-17
use latest travis

k-okada:

PR #107, opened 2015-02-23
Revert "use latest travis"

k-okada:
Reverts jsk-ros-pkg/jsk_pr2eus#106


k-okada:
this shouldn't be reverted, need to merge more latest travis #108

Issue #113, opened 2015-03-07
:angle-vector-duratoin does not support limb


garaemon:
`:angle-vector-duration` of `robot-interface` does not support limb and duration of `:angle-vector` method does not work correctly if we specify angle-vector for specified limb (controller).

garaemon:
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/117
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/114

These PR should solve this issue. @orikuma Please check if it works.

PR #114, opened 2015-03-07
[pr2eus] Support ctype in :angle-vector-duration

garaemon:
@aginika Please check if this works


aginika:
Ok, I will try in Baxter and PR2

garaemon:
If it does not work, please update this PR

garaemon:
Confirmed on PR2

garaemon:
Please be careful about this change

k-okada:
this will fixes #113

PR #117, opened 2015-03-13
fix :angle-vector-duration for limb controller, this will fix #113

### LINKED ISSUE #113: :angle-vector-duratoin does not support limb

k-okada:
- [test/pr2-ri-test-simple.{l,launch}] add test code for angle-vector-duration
- [robot-interfce.l] fix :angle-vector-duration for limb contorller


k-okada:
this is duplicates of #114

k-okada:
so, I'll use test only

Issue #120, opened 2015-03-18
send angle-vector to pr2 too slow


furushchev:
when using torso, moving arms are slow. I think this is because max joint vector of torso controller is too much slower than other controllers.

e.g.:

```lisp
irteusgl$ setq a (make-coords :pos #f(700 0 1300))
irteusgl$ send *pr2* :inverse-kinematics a :use-torso t
#f(242.932 60.0 74.0 70.0 -120.0 20.0 -30.0 180.0 3.13996 0.865655 -29.5923 -76.9422 -183.126 -77.7605 210.538 0.0 0.0)
irteusgl$ bench (send *ri* :angle-vector (send *pr2* :angle-vector) 2000)
irteusgl$ bench (send *ri* :angle-vector c 2000) ;; this assumes 2.0s to move body
;; time -> 25.3008[s] ;; this takes too long
#f(242.932 60.0 74.0 70.0 -120.0 20.0 -30.0 180.0 3.13996 0.865655 -29.5923 -76.9422 -183.126 -77.7605 210.538 0.0 0.0)
```

Any idea?

my idea is porting current pr2-interface's `:angle-vector` to `:angle-vector-safe` and implement `:angle-vector` like this:

```lisp 
 (:angle-vector
   (av &optional (tm 3000) &rest args)
     (send self :angle-vector-safe av tm (remove :torso-controller controller-type))))
```

garaemon:
scale optional argument is too large. we should use scale 1 to check if the
duration is too fast

2015年3月18日水曜日、Furushchev<notifications@github.com>さんは書きました:

> when using torso, moving arms are slow. I think this is because max joint
> vector of torso controller is too much slower than other controllers.
>
> e.g.:
>
> irteusgl$ setq a (make-coords :pos #f(700 0 1300))
> irteusgl$ send *pr2* :inverse-kinematics a :use-torso t#f(242.932 60.0 74.0 70.0 -120.0 20.0 -30.0 180.0 3.13996 0.865655 -29.5923 -76.9422 -183.126 -77.7605 210.538 0.0 0.0)
> irteusgl$ bench (send *ri* :angle-vector (send *pr2* :angle-vector) 2000)
[...]

PR #121, opened 2015-03-18
change default 5 to 1 as :scale in angle-vector

furushchev:
This change may affects on baxter interface.
So please change :scale of baxter's angle-vector to 5 @aginika 

ref: https://github.com/jsk-ros-pkg/jsk_pr2eus/issues/120

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -228,7 +228,7 @@
      warning))
 
   (:angle-vector-duration
-   (start end &optional (scale 5) (min-time 1.0) (ctype controller-type))
+   (start end &optional (scale 1) (min-time 1.0) (ctype controller-type))
    (let* ((unordered-joint-names 
            (flatten (mapcar #'(lambda (joint-param)
                                 (cdr (assoc :joint-names joint-param)))
@@ -246,7 +246,7 @@
          (let ((max-time (apply #'max time-list)))
            (max max-time min-time))))))
   (:angle-vector
-   (av &optional (tm nil) (ctype controller-type) (start-time 0) &key (scale 5) (min-time 1.0))
+   (av &optional (tm nil) (ctype controller-type) (start-time 0) &key (scale 1) (min-time 1.0))
    "tm (time to goal in msec)
    if designated tm is faster than fastest speed, use fastest speed
    if not specified, it will use 1/5 of the fastest speed .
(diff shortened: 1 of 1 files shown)


k-okada:
this may affects to all robots, not only pr2. plese check @snozawa @garaemon @mmurooka 

another solution is to change only for pr2.

furushchev:
I got approval of @garaemon .
Is there any serious problem with this change? > @snozawa @mmurooka

garaemon:
LGTM because
* staro -> all the programs are old and it does not have scale option
* hrp2 -> min-time=1.0 and scale=1 is enough

k-okada:
ok, so document should be updated too,
https://github.com/furushchev/jsk_pr2eus/blob/fast-default-scale/pr2eus/robot-interface.l#L252

also we need to confirm @orikuma, some day , staro will use latest
code, and jaxon will use latest code very soon. @YoheiKakiuchi
--
◉ Kei Okada

k-okada:
original discussion ->
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/86#issuecomment-69718776
--
◉ Kei Okada

furushchev:
updated documentation.

k-okada:
please be careful , everyone!

snozawa:
How about to use `[msec]` for return value of `:angle-vector-duration` and `:min-time` argument instead of `[sec]`?

In all lines using `:angle-vector-duration`, 
return values are multiplied by 1000 and change from `[sec]` to `[msec]`.
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/robot-interface.l#L258

Almost all Euslisp users are familiar with `[msec]` for duration (for example, `:angle-vector`), 
so using `[msec]` for duration is more simple and less confusing.

garaemon:
> How about to use [msec] for return value of :angle-vector-duration and :min-time argument instead of [sec]?

Just a historical reason why :angle-vector-duration returns time in sec. Hope somebody update it.

pr2eus/robot-interface.l
```diff
@@ -246,10 +246,10 @@
          (let ((max-time (apply #'max time-list)))
            (max max-time min-time))))))
   (:angle-vector
-   (av &optional (tm nil) (ctype controller-type) (start-time 0) &key (scale 5) (min-time 1.0))
+   (av &optional (tm nil) (ctype controller-type) (start-time 0) &key (scale 1) (min-time 1.0))
    "tm (time to goal in msec)
    if designated tm is faster than fastest speed, use fastest speed
-   if not specified, it will use 1/5 of the fastest speed .
+   if not specified, it will use 1/scale of the fastest speed .
    if :fastest is specefied use fastest speed calcurated from max speed"
```
k-okada: if not specifined, that is max speed, so we do not need :fastest anymore, is it correct?

https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/86#issuecomment-83123289

PR #124, opened 2015-03-24
[jsk_pr2eus] FIx :angle-vector-sequence by passing ctype argument to :angle-vector-duration

garaemon:
@orikuma Check it


garaemon:
Same patch to https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/114

PR #131, opened 2015-04-22
	[pr2eus] Fix :interpolatingp by using ros::*simple-goal-state-active* instead of actoinlib_msgs::GoalStatus::*active*

garaemon:
@orikuma


garaemon:
@aginika @furushchev Please check this on pr2

PR #138, opened 2015-05-14
[robot-interface.l] :angle-vector[-sequence] use default if nil ctype was

wkentaro:


k-okada:
this will not change any API, just to change behavior when you set nil to the controller-type

PR #141, opened 2015-05-14
[pr2eus] Check if already advertised using ros::get-topic-publisher

wkentaro:


wkentaro:
@k-okada Should I change something to have this PR be merged?

PR #148, opened 2015-06-23
[pr2eus] pass additional-weight-list when calling super class method

mmurooka:
This PR is necessary to do following
https://github.com/jsk-ros-pkg/jsk_model_tools/issues/119#issuecomment-114086317
```
2 use arm end coords but set small weight ( I suppose use greater number for eus arguments?) not to move
arms . Of cause, arm initial arm postures is set to grab object no the ground. From the kinematics point of 
view, this is almost equals to 1. So it should move.
```
@aginika


aginika:
@mmurooka Thanks a lots!

PR #154, opened 2015-07-29
[pr2eus] add real-time-simulation key

h-kamada:
angle-vector-simulation now consider execution duraction.

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -268,7 +268,7 @@
          (let ((max-time (apply #'max time-list)))
            (max max-time min-time))))))
   (:angle-vector-simulation
-   (av tm ctype)
+   (av tm ctype &optional (real-time-simulation nil))
    (let* ((prev-av (send robot :angle-vector))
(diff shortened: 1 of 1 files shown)


k-okada:
what is your expected behavior? if you comand `:angle-vector av 10000`, then the robot move its arm for 10 sec?
In this implementation the robot moves the arm quickly, for example 1sec and wait for 9 sec?

h-kamada:
> what is your expected behavior? if you comand :angle-vector av 10000, then the robot move its arm for 10 sec?
Yes.
> In this implementation the robot moves the arm quickly, for example 1sec and wait for 9 sec?
No.

Sleep is called in the dotimes loop.
So in every 100ms, program calculates the angle-vector and waits for 100ms pass.

Whole behavior is PR2 motion in irtviewer is updated in every 100ms.

h-kamada:
But if program sleep, `:angle-vector` doesn't return true until execution is done.
In real, `angle-vector` immediately return.
So, this commit may be bad idea.

PR #157, opened 2015-08-04
more realistic simulation mode

k-okada:
DO NOT MERGE YET

this is another implementation of #156 , please check if this work for you @h-kamada 

this is still under development, I just tested with following scripts
```
roseus pr2-interface.l "(pr2-init)" "(send *pr2* :arms :shoulder-p :joint-angle -30)" "(send *ri* :angle-vector-sequence (list (send *pr2* :reset-manip-pose) (send *pr2* :reset-pose)))" "(print \":angle-vector done\")"  "(print (send *ri* :wait-interpolation))"
```

TODO

- :wait-interpolation-smooth
- ~~:go-pos~~, go-velocity
- add test code
- cleanup code

LIMITATION

since this uses `*job-timer*` so you have to be top-level of explicitly call `:robot-interface-simulation-callback` to move robot arm,  so it may not work within smach, please check this @furushchev

### DIFF
--- a/pr2eus/pr2-interface.l
+++ b/pr2eus/pr2-interface.l
@@ -27,7 +27,9 @@
   :super robot-interface
   :slots (r-gripper-action l-gripper-action
                            move-base-action move-base-trajectory-action
-                           finger-pressure-origin))
+                           finger-pressure-origin
+                           current-goal-coords ;; for simulation-callback
+                           ))
 (defmethod pr2-interface
   (:init
    (&rest args &key (type :default-controller)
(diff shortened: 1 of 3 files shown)


h-kamada:
Ok. I'll do about this.

h-kamada:
worked in my env.

k-okada:
what is 'my env'? reizouko demo?

--
◉ Kei Okada

k-okada:
I haven't support :go-pose, :wait-interapolation-smooth, but your env did
not use these function?

--
◉ Kei Okada

h-kamada:
I don't mean it.
I checked this commit can do what I wanted to do in #156 .
Fridge demo use :go-pos function.

k-okada:
please check with fridge demo on euslisp

--
◉ Kei Okada

k-okada:
btw, if you run fridge demo on euslisp (without gazebo, real robot), I
think you saw a lot of
```
[ERROR] [1438838240.968802675]: "map" passed to lookupTransform argument
target_frame does not exist
```
message, but everything working well, is it correct? if so please write
issue for this. Displaying  error, but works good is confusing.

--
◉ Kei Okada

k-okada:
added :move-to

h-kamada:
> [ERROR] [1438838240.968802675]: "map" passed to lookupTransform argument target_frame does not exist ``` message, but everything working well, is it correct? if so please write issue for this. Displaying error, but works good is confusing.

Probably, this error message is by my mistake in jsk-ros-pkg/jsk_demos#1026 .

k-okada:
this has been supported in 
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/140 etc...

PR #158, opened 2015-08-06
[pr2eus] short fix

h-kamada:
Typo return-from word


k-okada:
nice, I added test code to catch this bug #159 
we hope that #159 will be fail and fixed with this pr

h-kamada:
I see. Thank you.

k-okada:
closed via #159

PR #167, opened 2015-08-26
Fix :move-to in sim mode (check frame-I'd) add test for :move-to

k-okada:
current implementation in simulationp is not correct, it always move to coordinates relative to current robot position, but that depends on `frame-id` arguments
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/d90a80e00a6aa03ad7fa632cf40551c13e2dc75b/pr2eus/pr2-interface.l#L394


k-okada:
DO NOT MERGE YET

PR #186, opened 2015-11-20
[.travis] update .travis to pass test with allow failure

furushchev:
now test always fails on maybe travis error

```bash
update complete.
+++wstool info --only=localname,uri
+++awk -v 'search=github.com/jsk-ros-pkg/jsk_pr2eus(|.git)$' 'BEGIN {FS=","} { if ($2 ~ search) print $1 }'
+++xargs -n1 wstool rm
Error: Too few arguments.
wstool remove [localname]*
xargs: wstool: exited with status 255; aborting
++++error
++++travis_time_end 31
++++set +x
traivs_time:end:4bqc9ws4:start=1447917760267563064,finish=1447917760853237304,duration=585674240

traivs_fold:end:setup_rosws
Function setup_rosws takes 0 min 0 sec
```


furushchev:
oh sorry, another pr is already open.

PR #189, opened 2015-11-23
[pr2eus/pr2.l] update pr2 model file

furushchev:
ref: https://github.com/jsk-ros-pkg/jsk_model_tools/pull/131#issuecomment-158950273

- [pr2eus/pr2.l] update pr2 model file
- [pr2eus/test/pr1012_camera_info.bag] update camera info
- [pr2eus/test/pr1012_sensors.bag] update sensor


k-okada:
Does this work on deb environment and source environment of jsk_model _tools, and I suppose if the answer is yes, we do not have problem like  jsk-ros-pkg/jsk_model_tools#131 (comment)

--
◉ Kei Okada

2015/11/24 0:47、Furushchev <notifications@github.com> のメッセージ:

furushchev:
I see. So is it ok to close? @YoheiKakiuchi

furushchev:
ping

k-okada:
since we will not have newer deb version of jsk_model_tools, so we'd better
to keep old pr2.l model until we upgrade the robot to indigo

http://www.ros.org/debbuild/hydro.html?q=euscollada

--
◉ Kei Okada

furushchev:
Closed. If there is demand, I'll reopen.

PR #191, opened 2015-11-30
[pr2eus/robot-interface.l, pr2eus/pr2-interface.l] fix: :wait-interpolation returns :interpolatingp on real robot

furushchev:
**NOTICE**
This PR changes return value of `:wait-interpolation`

**SUMMARY**
the return value of `:wait-interpolation` is defined in docs as below:

> return values is a list of interpolatingp for all controllers, so (null (some #'identity (send *ri* :wait-interpolation))) -> t if all interpolation has stopped

but on real robot it returns `(send-all controller-actions :wait-for-result))`
So I fixed according with docs.

tested on pr2-simulator and pr1012 real robot.

discussed at https://github.com/jsk-ros-pkg/jsk_pr2eus/issues/187

### DIFF
--- a/pr2eus/pr2-interface.l
+++ b/pr2eus/pr2-interface.l
@@ -113,24 +113,18 @@
   (:publish-joint-state () ;; overwrite for pr2
    (send-super :publish-joint-state (append (send robot :joint-list) (send robot :larm :gripper :joint-list) (send robot :rarm :gripper :joint-list))))
   ;;
-  (:wait-interpolation (&rest args);; overwrite for pr2, due to some joint is stll moving after joint-trajectry-action stops
+  (:wait-interpolation (&optional (ctype) (timeout 0)) ;; overwrite for pr2, due to some joint is stll moving after joint-trajectry-action stops
    (when (send self :simulation-modep) (return-from :wait-interpolation (send-super :wait-interpolation)))
-   (let (result)
-     (ros::ros-info "wait-interpolation debug: start")
-   ;;  (setq result (send-all controller-actions :wait-for-result))
-     ;;(setq result (send-super* :wait-interpolation args))
-     (dolist (ca controller-actions)
-       (push (send ca :wait-for-result) result))
-     (ros::ros-info "wait-interpolation debug: end")
-     (setq result (reverse result))
-	 (while (ros::ok)
-	   (send self :update-robot-state)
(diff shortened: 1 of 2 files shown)


furushchev:
@k-okada @snozawa How do you think of it?

snozawa:
I think we also need to discuss the return value of `:interpolatingp`.

According this discussion,
https://github.com/jsk-ros-pkg/jsk_pr2eus/issues/187#issuecomment-160243097
https://github.com/jsk-ros-pkg/jsk_pr2eus/issues/187#issuecomment-160243678
do we need to return `nil` when interpolation is finished and some error occurs.

Questions:
- Should return values of `:wait-interpolation` and `:interpolatingp`  be same?
- `:interpolationgp` returns `t` if interpolation finished normally, otherwise returns `nil` (not finished or finished under some errors). Is this correct?

pr2eus/pr2-interface.l
```diff
@@ -113,24 +113,18 @@
...
-     (setq result (reverse result))
-	 (while (ros::ok)
-	   (send self :update-robot-state)
-	   (when (every #'(lambda(x)(< (abs (send x :joint-velocity))
-				       (if (derivedp x rotational-joint) 0.05 0.001)))
-			  (send robot :joint-list))
-		 (return)))
-	 result))
-
+   (ros::ros-info "wait-interpolation debug: start")
```
snozawa: For minor comments, it would be better to add more information for this like
`wait-interpolation debug: start of robot-interface's :wait-interpolation`

furushchev:
@snozawa Thank you for revieweing.
> - Should return values of :wait-interpolation and :interpolatingp be same?

No, 
- `robot-interface` and `action-controller` instances both have each `:interpolatingp` method
- `:interpolatingp` of `robot-interface` returns `t` **only** when all controllers are not interpolating, otherwise returns `nil`.
- `:wait-interpolation` of `robot-interface` returns list of  results of all controllers' `:interpolatingp`.
so when robot stops,

```
:wait-interpolation => (nil nil nil nil)
:interpolating => nil
```

> - :interpolationgp returns t if interpolation finished normally, otherwise returns nil (not finished or finished under some errors). Is this correct?

I think it returns `t` only when finished any motion.
When "finished under some errors", it should return `nil`.

snozawa:
Thanks, I see.
LGTM.

PR #192, opened 2015-12-01
[pr2eus/speak.l] add default variable for waiting speak

furushchev:
**SUMMARY**

- by setting `t`/`nil` for variable `*speak-wait*`, we can change default behavior of speak function

**EFFECT**

global variable `*speak-wait*` is defined.

### DIFF
--- a/pr2eus/speak.l
+++ b/pr2eus/speak.l
@@ -3,7 +3,9 @@
 
 (ros::load-ros-manifest "sound_play")
 
-(defun send-speak-msg (msg &key (topic-name "robotsound") (timeout 0) wait)
+(defparameter *speak-wait* nil)
+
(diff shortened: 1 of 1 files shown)


furushchev:
最近多い気がします。。

```
ERROR: cannot download default sources list from:

https://raw.github.com/ros/rosdistro/master/rosdep/sources.list.d/20-default.list

Website may be down.
```

furushchev:
ok now test passed.

k-okada:
when we should wait and when not wait.

furushchev:
`:wait` option is needed when, for example, robot speaks on what he **will** do (e.g. "I turn left.") or question which need to be answered by people, which are order sensitive.
But in debugging or just tweeting we does not need to wait.

PR #196, opened 2015-12-09
[pr2eus] add :go-waitp

h-kamada:
* check whether PR2 is moving
* In simulation mode, just return nil

### DIFF
--- a/pr2eus/pr2-interface.l
+++ b/pr2eus/pr2-interface.l
@@ -516,6 +516,20 @@
          ))
      (setq move-base-goal-msg nil) ;; :move-to-wait has been called
      ret))
+
+  (:go-waitp
+   ()
+   (send self :spin-once) ;; update
+   (when (send self :simulation-modep)
+     (return-from :go-waitp nil)) ;; temporary
+   (cond
+    ((or
+      (equal (send move-base-action :get-state) ros::*simple-goal-state-active*)
+      (equal (send move-base-trajectory-action :get-state) ros::*simple-goal-state-active*))
+     (return-from :go-waitp t))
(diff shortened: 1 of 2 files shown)


k-okada:
add test code at https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/test/pr2-ri-test-simple.l#L52, just call :go-waitp before :go-wait is ok for the first testcode.

h-kamada:
But, in simulation mode, `:go-pos-wait ` is calling `:robot-interface-simulation-callback` and, in calback, robot coordinates are just changed. (not watching about actionlib status)
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/pr2-interface.l#L745-L760

So, currently `:go-waitp` test in simulation seems useless.

k-okada:
yes ,but just to check if you have something wrong in the code, like typo
or undefined variables or...

--
◉ Kei Okada

h-kamada:
OK, fixed. Currently `:go-waitp` just return nil in simulation mode.

garaemon:
should be `:go-wait-p`

k-okada:
for historical reasons, we use :wait-interpolation and :interpolatingp,
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/robot-interface.l#L490
so I suggested to use :go-waitp , but may be it was not a good idea.

so
:go-pos
:go-wait
,,
so
:go-waitp
:gop
:go-wait-p
:go-p
:goingp
...

:go-wait-p may better?


--
◉ Kei Okada

PR #209, opened 2016-02-19
[pr2eus] add get-grasp-result

h-kamada:
- add get-grasp-result method
- add :wait key to start-grasp method


k-okada:
please check if finger joint of robot model in *ri* is updated, when we
call :state

--
◉ Kei Okada

h-kamada:
In robot-interface class, :state method calls :update-robot-state method. This method calls :spin-once method, so I think gripper joint is updated.

k-okada:
if gripper joint  is updated, you do not have to add newer callback
function, just use (*ri* . robot) to obtain the distance between the fingers

--
◉ Kei Okada

PR #218, opened 2016-04-18
[pr2eus/robot-interface.l] wait /clock publish for a while when /use_sim_time is true

furushchev:
exiting process checking `/clock` is published only once is too severe for not high spec machine.


pr2eus/robot-interface.l
```diff
@@ -125,18 +125,25 @@
...
-     (let ((cntr 0))
-       (while (and (= (ros::get-num-publishers "/clock") 0)
-                   (<= cntr 10))
-         (unix::usleep (* 100 1000)) ;; 100ms
-         (incf cntr)))
-     (when (= (ros::get-num-publishers "/clock") 0)
-       (ros::ros-error "/use_sim_time is TRUE and /clock is NOT PUBLISHED")
-       (exit 1))
-     (ros::unsubscribe "/clock"))
+   (let ((start-time (now))
```
k-okada: where do we define  'now ' ?

furushchev: defined in https://github.com/euslisp/EusLisp/blob/master/lib/llib/time.l#L304

PR #219, opened 2016-04-21
update .travis

k-okada:


k-okada:
need to use 0.4.5 -> https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/221

Issue #206, opened 2016-02-03
pr2 arm-object collision check


iKrishneel:
I want the pr2 arm not to collide with the table during manipulation and I tried to use `pqp-collision-check` method. 

The problem is if the links of the arm is inside the cube (representing the table), collision check does not report collision as shown in the attached picture. However if the arm is on the edge of the cube,  collision is reported. 

[Code for this is here: ](https://gist.github.com/iKrishneel/6b5df868b11794cc5aa6)

How to work about this issue?

![screenshot from 2016-02-03 19 19 29](https://cloud.githubusercontent.com/assets/9315167/12788911/e6423278-cade-11e5-821e-ccb8a9dcb62c.png)

k-okada:
nice catch, this is very complex problem which caused from conversion between urdf model and eus model. urdf model consists of several links  with rotational joint and fixed joint whereas eus model originally assume a link is consists of one link, not a multiple link connected by fixed joint as urdf do.

see https://gist.github.com/k-okada/2d0e4eacd2f0f5278fa7 for templary workaround (and also i modifed for better visualization, please compare this with yours carefully)

For better solution, we'll continue discussion at -> 
https://github.com/jsk-ros-pkg/jsk_model_tools/issues/171

chiwunau:
I have a similar problem, where I am trying to solve IK while avoiding the robot's arm and base collide with the door model. The screenshot below shows the result when I put every pair of door links and links from ```(send *pr2* :link-list (send *pr2* :rarm :end-coords :parent))``` to ```:collision-avoidance-link-pair``` of IK. ([code](https://github.com/chiwunau/pr2_open_b2_door_demo/blob/contact-search/euslisp/contact_state/search-motion.l#L29))
It seem that when searching in [null-space](https://github.com/euslisp/jskeus/blob/master/irteus/demo/null-space-ik.l), It only consider collision of bodies in the link-list. In this case many parts of arm and base are missing.
![Uploading Screenshot_from_2016-05-24 19:16:44.png…]()

k-okada:
@chiwunau @iKrishneel 
can we try "workaround" for this issue?
1) create list contains "body object" for each links, manually
2) create convex hull of these objects
3) replace collision object for each links
4) write this procedure within pr2-utils.l
@furushchev can you help them?

chiwunau:
Thank you, I will get @furushchev help and try writing the code.

chiwunau:
[...]

PR #232, opened 2016-06-14
Include redundant links when calculate collision

chiwunau:
This is to fix the problem discussed in https://github.com/jsk-ros-pkg/jsk_pr2eus/issues/206

### DIFF
--- a/pr2eus/pr2-utils.l
+++ b/pr2eus/pr2-utils.l
@@ -3,6 +3,42 @@
 ;;
 (require :pr2 "package://pr2eus/pr2.l")
 
+;;Take account of redundant links when calculate collision
+(defmethod euscollada-robot
+  (:make-collision-model-for-links
+   (&key (fat 0) (collision-func 'pqp-collision-check) ((:links ls) (send self :links)))
+
+
+   (dolist (ll ls)
+     (send ll
+           (read-from-string
+            (format nil ":make-~Amodel"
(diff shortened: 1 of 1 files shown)


k-okada:
@h-kamada please check if this PR is OK for you.

h-kamada:
The code itself seems ok.

I faced the collision problem when I was writing about `:start-grasp` , `stop-grasp` method of `*pr2*`.
Currently  `:start-grasp` is checking grasp by checking collisions of all gripper links. ( --> https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/pr2-utils.l#L263-L272)

I forgot details about this , but  I think the problem was that `pqp-collision-check` didn't work well if I used `(pqp-collision-check *pr2* target)`.
So I tried to check whether I can use `(pqp-collision-check *pr2* target)` if I use this code.
But the original code also worked well (not knowing why). 
So I couldn't check whether this code is ok for my problem.

chiwunau:
In this PR (may be there is better solution), collision models are basically attached to their ascendant link found in eus model link-list, so gripper link consider as wrist_roll_link.
which means ```(pqp-collision-check (elt (send *pr2* :larm :links) 6) cube)``` returns 1 in the picture below.
So the original code may be needed to specify only gripper.
screenshot_from_2016-06-15 19 24 39

k-okada:
@snozawa , @YoheiKakiuchi  could take a look at this PR and check if we can import this code into irteus level

PR #236, opened 2016-06-20
wait for new feedback message arrival for :wait-interpolation-smooth

k-okada:
@pazeshun this may solve your problem, sometimes `:spin-once` in `:wait-interpolation` (or `:interpolating-smoothp` more precise) did not catch new feedback mesasge, and use old one,


pazeshun:
Thanks. I want to test this, but I have no time today...

pazeshun:
`:wait-interpolation-smooth` succeeded with this PR.

PR #238, opened 2016-06-25
add test-state-wait-until-update

k-okada:
sometines :state .. :wait-unitl-update t did not returns

c.f.: see test in https://github.com/jsk-ros-pkg/jsk_robot/pull/627


k-okada:
Merged, this is only for simulation

PR #240, opened 2016-06-29
fix bug in robot-interface.l

chiwunau:
this bug lead to error when using ```(send *ri* :move-trajectory-sequence :send-action t)```

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -1556,7 +1556,7 @@ send-action [ send message to action server, it means robot will move ]"
          (send msg :points (list pt1 pt2)))
        ))
      (send goal :goal :trajectory msg)
-     (when send-action and move-base-trajectory-action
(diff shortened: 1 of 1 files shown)


chiwunau:
done in https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/237

k-okada:
sorry I missed this PR, 
1) could you explain more about when you need to use :send-action option? which is disable for default. https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/240/commits/af48e627af02b45c16c7ec37691cd0ce9cfa8e81#diff-15f6920b924513f6027d36b361f85bbdR1475
2) could you add test code that fails without this PR?

Cc: @furushchev

k-okada:
@chiwunau can we close this? I think it has been solved at https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/237/files#r76525642

PR #241, opened 2016-07-01
[pr2eus/speak.l] use single speak-action-client

furushchev:
prevents creating action client everytime speaking


furushchev:
@k-okada Please merge if there is no problem.

k-okada:
Is ther any explicit problem if you create multiple actionlib client

furushchev:
@k-okada we cannot send cancel to server to stop in loop play mode. (in existing sound_play node, we cannot notice that by chance because of bug)
Also duplicating action server occurs warning messages.

PR #242, opened 2016-07-16
[pr2eus/pr2.l] comment out pr2 function for pr2-robot

furushchev:
`pr2` function is overdriven by initializer of `pr2-sensor-robot` at the below of `pr2eus/pr2.l`.
Duplicate declaration of function with the same name prevents compilation.


k-okada:
Not to confused, how about

 +#| ;; overdriven by pr2-sensor-robot, defined below or defined line xxx of this file, they are generated from https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/make-pr2-model-file.l#L37

pr2eus/pr2.l
```diff
@@ -125,7 +125,11 @@
      ))
   )
 ;;
+#|
+;; overdriven by pr2-sensor-robot, which is defined below of this code.
+;; (auto generated from make-pr2-model-file.l)
```
furushchev: @k-okada Updated!

PR #252, opened 2016-11-01
fix robot-moveit.l

k-okada:
@YoheiKakiuchi, please review. This PR incldues several changes in robot motion

1. :angle-vector-motion-plan send trajectory not only joints incldues move-arm
2. :angle-vector-motion-plan has new arguments, named :total time, which scales planned motion duration
3. :add :start-offset-time, which adde start offet to trajecotry (for fetch)

### DIFF
--- a/pr2eus_moveit/euslisp/robot-moveit.l
+++ b/pr2eus_moveit/euslisp/robot-moveit.l
@@ -504,21 +504,23 @@
        )))
   (:trajectory-filter ;; simple trajectory for zero duration
    (traj &key (copy) (total-time 5000.0) (minimum-time 0.001) (start-offset-time) &rest args &allow-other-keys)
+   (let ((orig-total-time (send (send (car (last (send traj :points))) :time_from_start) :to-sec)))
    (when (and minimum-time (not start-offset-time)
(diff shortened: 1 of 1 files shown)


pr2eus_moveit/euslisp/robot-moveit.l
```diff
@@ -463,18 +463,44 @@
...
      (setq ret (send* self :angle-vector-make-trajectory av args))
      (when ret
        (setq traj (send ret :trajectory :joint_trajectory))
-       (when (< (send (send (car (last (send traj :points))) :time_from_start) :to-sec) 0.001)
+       (setq orig-total-time (send (send (car (last (send traj :points))) :time_from_start) :to-sec))
+       (when (< orig-total-time 0.001)
          (unless reset-total-time
            (ros::ros-error "Trajectory has very short duration")
            (return-from :angle-vector-motion-plan nil))
          (ros::ros-warn "reset Trajectory Total time")
```
k-okada: @YoheiKakiuchi Do we need `reset-total-time` feature, if we have https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/252/files#diff-113ce1b47d7cec61152334f1050188c1R479 feature? which always try to scale planned trajectory to use setting 'tm' value?

PR #272, opened 2016-12-09
use link-list instead of (car link-list) in use-base condition

chiwunau:
This change makes dual-arm iK with use-base work.

If there is no special reason to use (car ll).
@furushchev @h-kamada 

dual-arm IK with use-base:
```
(setq c1 (send *pr2* :larm :end-coords :copy-worldcoords))
(send c1 :translate #f(300 0 0) :world)
(setq c2 (send *pr2* :rarm :end-coords :copy-worldcoords))
(send c2 :translate #f(300 0 0) :world)

(send *pr2* :inverse-kinematics (list  c1
                                      c2)
      :move-target (list  (send *pr2* :larm :end-coords)
                         (send *pr2* :rarm :end-coords))
      :link-list (list (send *pr2* :link-list (send *pr2* :larm :end-coords :parent))
                       (send *pr2* :link-list (send *pr2* :rarm :end-coords :parent)))
      :rotation-axis (list nil nil)
      :base-range (list :min (float-vector -10000 -10000 -80)
                        :max (float-vector 10000 10000  80))

      :use-base t
      :debug-view :no-message)
```


furushchev:
@chiwunau Great work! Could you write a test code for that?

k-okada:
ping @chiwunau

chiwunau:
test code added, please merge if no problem.

mmurooka:
Please wait merge.
@furushchev  said that this does not work correct with other robot.
@furushchev, cloud you check this for PR2.

mmurooka:
> Please wait merge.
@furushchev said that this does not work correct with other robot.
@furushchev, cloud you check this for PR2.

was mistaken communication and this PR is OK.sorry for confusion.

k-okada:
@Affonso-Gui this is an example of dual-arm manipulation, you can also use base if you want increase workspace. I'll merge and release this PR with in this week.

Affonso-Gui:
Great! Good to see another example of dual-arm ik, and :use-base will be very useful when adapting the code for smaller robots.

PR #276, opened 2016-12-27
[pr2eus_moveit] pass start-offset-time as starttime to :send-trajectory

knorth55:
solve: https://github.com/start-jsk/jsk_apc/issues/1989
this cause baxter arm jerky arm motion.
`send-trajectory` `starttime` should be equal to `angle-vector-motion-plan`'s' `start-offset-time` (according to `angle-vector` https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/robot-interface.l#L395)


knorth55:
restart travis and build passed!

PR #291, opened 2017-05-09
[pr2] Fix &rest after &key parameters

Affonso-Gui:
Fix &rest after &key parameters at pr2-interface.l

### DIFF
--- a/pr2eus/pr2-interface.l
+++ b/pr2eus/pr2-interface.l
@@ -407,7 +407,7 @@ Example: (send self :gripper :rarm :position) => 0.00"
 
 (defmethod pr2-interface
   (:angle-vector-with-constraint
-   (av1 &optional (tm 3000) (arm :arms) &key (rotation-axis t) (translation-axis t) &rest args)
+   (av1 &optional (tm 3000) (arm :arms) &key (rotation-axis t) (translation-axis t) &allow-other-keys)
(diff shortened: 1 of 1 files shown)


pr2eus/pr2-interface.l
```diff
@@ -407,7 +407,7 @@ Example: (send self :gripper :rarm :position) => 0.00"
 
 (defmethod pr2-interface
   (:angle-vector-with-constraint
-   (av1 &optional (tm 3000) (arm :arms) &key (rotation-axis t) (translation-axis t) &rest args)
+   (av1 &optional (tm 3000) (arm :arms) &key (rotation-axis t) (translation-axis t) &allow-other-keys)
```
furushchev: 1. `&allow-other-keys` only works with `&rest`, so it is wrong to delete `&rest`.
1. Currently variable `args` is not used anywhere, but this function works so I think we don't have to change if you don't have any trouble on this.

Affonso-Gui: There is no meaning in defining `&rest` arguments after `&key` declaration, since sending anything after the keys will led to errors like `no such keyword` or `keyword expected`. Never being reached, args value will always be nil.

If the purpose of `&rest` here is to allow the user to pass undefined arguments without errors (since args is not used anywhere), the use of `&allow-other-keys` should be correct. If not, would be better to erase `&rest` altogether, avoiding misleading coding.

PR #296, opened 2017-05-22
.travis.yml: re-enable pr2-ri-test (using gazebo) for indigo

k-okada:
was disabled in #25


k-okada:
@furushchev it's easy!

PR #301, opened 2017-06-21
use make-caemra-from-ros-camera-info-aux inroseus, in order to generate pr2 model corresponding to jsk-ros-pkg/jsk_roseus#526

k-okada:
rewrited version of #300

- remove make-camera-from-ros-camera-info-aux, it is defined in roseus 1.0.1
- [pr2eus/pr2.l] update make-camera-from-ros-camera-info-aux


k-okada:
i see, so why we define make-camera.... here, to work without roseus?

k-okada:
ok, fixed test code, just to pass camera test only on hydro @YoheiKakiuchi

YoheiKakiuchi:
LGTM

PR #315, opened 2017-07-13
[pr2eus_moveit] add 0.5 seconds sleep after collision object pub

knorth55:
According to moveit tutorials, we need to put sleep after topic publishing.
I have several reports that it sometimes fails to wipe all scenes.

http://docs.ros.org/indigo/api/pr2_moveit_tutorials/html/planning/src/doc/planning_scene_ros_api_tutorial.html


knorth55:
adding `unix::usleep` is not smart solution, but I couldn't find out better solution.

k-okada:
please add following comment within source code
```
According to moveit tutorials, we need to put sleep after topic publishing.
http://docs.ros.org/indigo/api/pr2_moveit_tutorials/html/planning/src/doc/planning_scene_ros_api_tutorial.html
adding unix::usleep is not smart solution, but I couldn't find out better
solution.
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/315#issuecomment-315015718
```

--
◉ Kei Okada


2017-07-13 17:54 GMT+09:00 Shingo Kitagawa <notifications@github.com>:

k-okada:
added comment

knorth55:
Thank you!

PR #318, opened 2017-09-09
pr2eus: add text-to-spech method to robot-interface

furushchev:
Also addressed in https://github.com/jsk-ros-pkg/jsk_robot/pull/833

This PR includes:
- Cleanup unused text-to-speech methods
- Add text-to-speech method caller to robot-interface

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -1039,6 +1039,7 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
    "get information about gripper"
    (error "This method is responsible to sub class~%"))
   ) ;; robot-interface
+
 ;; ros visualization methods
 (defmethod robot-interface
   (:joint-trajectory-to-angle-vector-list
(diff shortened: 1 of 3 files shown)


furushchev:
Test passed. Confirmed to work on real robot (PR1012, fetch).

k-okada:
@furushchev does google translate is not working? please also check https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/263, one idea is to support 'text' argument of ':speak' supports both 'string' and '((:jp . "こんにちは") (:en . "Hello"))'

furushchev:
@k-okada Thank you very much for feedback!

> does google translate is not working?

Not working. Recently google updated their authentication policy to use request with correct header and token but gstreamer (on which `sound_play` depends) unfortunately does not support this feature.
(We still have alternative method to use google api if we use the package I made several months ago https://github.com/furushchev/ros_gtts. Is it ok to send a pull request to `jsk_3rdparty`?)

> one idea is to support 'text' argument of ':speak' supports both 'string' and '((:jp . "こんにちは") (:en . "Hello"))'

That's a good idea! One possible problem is that we cannot know which language is available at euslisp level.
IMO, having global variable `*speak-language*` and set one language as its value solves this problem, but I don't think it is elegant so much...

furushchev:
@k-okada @snozawa Any advice or comment?

pr2eus/speak.l
```diff
@@ -8,7 +8,8 @@
 (defparameter *speak-timeout* 20)
 
 (defun send-speak-msg (msg
-                       &key (topic-name "robotsound") (timeout *speak-timeout*)
+                       &key (topic-name "robotsound")
+                            (timeout *speak-timeout*)
```
k-okada: this change may not necessary.

pr2eus/robot-interface.l
```diff
@@ -1186,6 +1187,26 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
...
+     (:en (send* self :speak-en text (remprop :lang args)))
+     (:ja (send* self :speak-jp text (remprop :lang args)))
+     (t (error "Unsupported language ~A" lang))))
+  (:speak-en (text &rest args)
+    "Speak english sentence"
+    (apply #'speak-en text args))
+  (:speak-jp (text &rest args)
+    "Speak japanese sentence"
+    (apply #'speak-jp text args))
+) ;; defmethod robot-interface (text-to-speech)
```
k-okada: how about add `:(send-speak-msg` here and change
```
(defun speak-jp (text &rest args)
 (warning-message 3 "Function `speak-jp` will be deprecated, please use (send *ri* :speak \"text\" :lang :ja") 
 (send* *ri* :speak text :lang :ja args) ;; this is dangerous, because it assumes *ri*..... is this ok? if not leave current implementation here, just to show warning message"

furushchev: @k-okada Sorry for a late response.
That's indeed a good idea. I updated as followings:
- added warning message for deprecated functions `speak-jp` and `speak-en`.
- move text-to-speech into `robot-interface`
- call `robot-interface` method `:speak-jp` internally in `speak-jp` function. (same for `speak-en`)
- updated test codes

k-okada:
confirmed by @708yamaguchi  https://github.com/jsk-ros-pkg/jsk_robot/pull/866#issuecomment-342725344

PR #320, opened 2017-10-01
from kinetic load-ros-manifest packages without srv/msg is deprecated

k-okada:


furushchev:
@k-okada This fix is included in https://github.com/jsk-ros-pkg/jsk_pr2eus/commit/01b5e0fbb6840c0e3be93d0423809337ffeccf54 and was already merged into master.

PR #324, opened 2017-10-11
Use service call for collision-object-publisher

knorth55:
Improve version of #317 (thanks to @eisoku9618)

Use `ApplyPlanningScene` service to operate `collision_objects` and `attached_collision_objects`.
Currently, `publish and sleep` method is implemented, but the old-style method is not recommended. (https://github.com/ros-planning/moveit_msgs/pull/40#issuecomment-335734542)
We use `service-call` method instead.


eisoku9618:
Thank you! You can squash my commit and your first commit if you like.

pr2eus_moveit/euslisp/collision-object-publisher.l
```diff
@@ -174,52 +178,62 @@
...
-         (send msg :object :operation moveit_msgs::CollisionObject::*REMOVE*)
-         (ros::publish attached-topicname msg)
-         ;; According to moveit tutorials, we need to put sleep after topic publishing.
-         ;; http://docs.ros.org/indigo/api/pr2_moveit_tutorials/html/planning/src/doc/planning_scene_ros_api_tutorial.html
-         ;; adding unix::usleep is not smart solution, but I couldn't find out better solution.
-         ;; https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/315#issuecomment-315015718
-         (unix::usleep (* 500 1000)))
+       (send req :scene :robot_state :attached_collision_objects
+             (mapcar #'(lambda (msg)
+                         (progn
```
YoheiKakiuchi: progn is not required inside lamda

knorth55: thank you so much! I updated.

knorth55:
I checked with baxter and it works with no problem.

pr2eus_moveit/euslisp/collision-object-publisher.l
YoheiKakiuchi: +1

PR #328, opened 2017-11-01
travis: drop hydro

furushchev:


furushchev:
test failed on 3 conditions but they also failed on the previous builds so it's ok to be merged.

travis_pr2eus

PR #331, opened 2017-12-06
pr2eus: partially revert speak function

furushchev:
This pull request reverts #318 + integrate TTS methods to robot-interface

PR #332, opened 2017-12-11
pr2eus: partially rever speak function

k-okada:
revrited version of #331, we need `(speak-jp)` withiht `*ri*`

- sound_play could not run within travis/jenkins, so use hztest aso dummy subscriber
- pr2eus: integrate speak function
- Revert "pr2eus: add text-to-spech method to robot-interface (#318)"
  This reverts commit ecb2a1e29d2d56ae16035064c91617f2b0afa786.

### DIFF
--- a/pr2eus/speak.l
+++ b/pr2eus/speak.l
@@ -3,48 +3,76 @@
 
 (ros::load-ros-manifest "sound_play")
 
-(defun speak-jp (text &key wait timeout (topic-name "robotsound_jp"))
-  "Speak japanese sentence using text-to-speech service.
+(defparameter *speak-wait* nil)
+(defparameter *speak-action-clients* (make-hash-table))
+(defparameter *speak-timeout* 20)
 
-   Args:
-    - text: sentence to speak
-    - topic-name: topic name space for sound_play server
-    - wait: wait the end of speech if enabled
-    - timeout: timeout for waiting, this is valid only if wait is enabled"
+(defun send-speak-msg (msg
+                       &key (topic-name "robotsound") (timeout *speak-timeout*)
+                            (wait *speak-wait*))
+  (cond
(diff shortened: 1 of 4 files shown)


pr2eus/speak.l
```diff
@@ -3,48 +3,72 @@
...
-
-  (let ((timeout-bak (send *ri* :speak-timeout)))
-    (when timeout
-      (send *ri* :speak-timeout timeout))
-    (prog1
-        (send *ri* :speak-jp text :wait wait :topic-name topic-name)
-      (when timeout
-        (send *ri* :speak-timeout timeout-bak)))))
-
-(defun speak-en (text &key wait timeout (topic-name "robotsound"))
```
k-okada: @furushchev , are we ok to remove 'timeout' keyword?

pr2eus/speak.l
```diff
@@ -3,48 +3,72 @@
...
-      (when timeout
-        (send *ri* :speak-timeout timeout-bak)))))
+     text: sentence to speak
+     lang: language to speak, currently :en or :ja are supported.
+     topic-name: namespace of sound_play node
+     wait: wait the end of speech if enabled"
+  (play-sound text
+              :topic-name topic-name
+              :wait wait
+              :arg2 (if (keywordp lang)
```
k-okada: what is `arg2` arguments? are we still using this?

pr2eus/speak.l
furushchev: @k-okada Yes, I changed to use `topic` if no response returns from action server, instead of waiting for it.

pr2eus/speak.l
furushchev: It is from `sound_play` message. http://docs.ros.org/jade/api/sound_play/html/msg/SoundRequest.html
We are using for specifying language or speech engine.

PR #334, opened 2018-01-24
robot-interface: upgrade service name for change-inflation-range

furushchev:


k-okada:
@furushchev please add link to the commit, which requires this change.

furushchev:
@k-okada Sorry for sending PR without description. This pull request is in progress. Please wait for update.

PR #338, opened 2018-04-02
increase stall_velocity_threshold to pass grasp test

k-okada:
Since pr2_gazebo for kinetic was released early 2018, travis failed in kinetic test (https://travis-ci.org/jsk-ros-pkg/jsk_pr2eus/builds/361065737, https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/337)

It seems `start-grasp` never returns because of weak effort? or never detected stall situation.
This might be not a good solution, we need to more better simualtion of gripper. For example we set 5000 effort for simulation, but real robot uses ~25.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ python:
   - "2.7"
 compiler:
   - gcc
+services:
+  - docker
 notifications:
   email:
     on_success: always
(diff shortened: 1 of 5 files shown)


furushchev:
Did you use the description in version after https://github.com/PR2/pr2_common/pull/255 ?

furushchev:
@k-okada FYI, many techniques (or hacks) were needed till PR2 successfully fetched cans from the fridge.
I'm not obvious if all of these improvements or fixes were released as apt packages.

k-okada:
@furushchev thank you for info
from https://api.travis-ci.org/v3/job/361065740/log.txt, we are using 1.12.1, which should include your fix
```
ros-kinetic-pr2-description is already the newest version (1.12.1-0xenial-20180328-052704-0800).
```

PR #352, opened 2018-06-12
pr2-ri-test.l : add test to check :wait-interpolation, with timeout

k-okada:
to check return value of :wait-interpolation with timeout(see  #351)


pr2eus/robot-interface.l
```diff
@@ -81,7 +81,12 @@
           (setq current-time 0)
           )
         av)))
-  (:interpolatingp () (not (null timer-sequence)))
+  (:interpolatingp
+   ()
+   (cond ((send ri :simulation-modep)
+          (not (null timer-sequence)))
+         (t ;; real robot
+          (eq (send self :get-state) ros::*simple-goal-state-active*))))
```
furushchev: According to my review at #351, how about changing this line to:

`(not (eq (send self :get-state) ros::*simple-goal-state-done*))`?

c.f. https://github.com/jsk-ros-pkg/jsk_roseus/blob/master/roseus/euslisp/actionlib.l#L203

k-okada: updated, note that `get self :get-state` do not return `ros::*simple-goal-state-active*`, need to check `ros::simple-state`

furushchev:
@k-okada 
I need this PR for make interruption work.
I added minor fix to your branch.
https://github.com/k-okada/jsk_pr2eus/pull/8
Please consider to merge this PR. :+1:

k-okada:
@furushchev so can we merge this? and also could you summarize our discussion ?

furushchev:
@k-okada Although the test failed, this PR looks good to me :+1: 

It is because the test assumes the difference between two coords is `0.0` in float precision, it can be relaxed?


```xml
<testsuite tests="6" failures="1" disabled="0" errors="0" time="0" name="AllTests">

  <testcase name="all-test" status="run" time="0" classname="all-test">

  </testcase>

  <testcase name="test-angle-vector" status="run" time="0" classname="test-angle-vector">

  </testcase>

  <testcase name="test-link" status="run" time="0" classname="test-link">

   <failure message="check link position #&lt;bodyset-link #X75ccfb0 head_tilt_link  0.93 0.0 1222.125 / 0.0 0.0 0.0&gt; #&lt;bodyset-link #X794d910 head_tilt_link  0.93 0.0 1222.125 / 0.0 0.0 0.0&gt;" type="AssertionError">

Test:(= (norm (send (send link-unstable :worldcoords) :difference-position (send link-stable :worldcoords))) 0.0)

Trace:"eus/rosunit-make_pr2_model_file_test.xml

).^[0m

0.0&amp;gt; ^[0mmstart testing [test-link]

vector]
```

furushchev:
@k-okada Thanks!

PR #356, opened 2018-07-13
Generalize :wait-interpolation-smooth

pazeshun:
Enable `:wait-interpolation-smooth` to support all cases in https://github.com/start-jsk/rtmros_common/pull/1052#issuecomment-404753185
- Exit from the method at the correct timing even if feedback from joint trajectory action is not published
- Don't relay on `time_from_start` of feedback because the behaviour varies by robot


pazeshun:
Tested with...
- PR2
- Baxter (with gripper)
- Fetch
- HIRONX (simulation)

PR #359, opened 2018-07-25
remove unnecessary start-offset-time

knorth55:
`start-offset-time` is set in `:send-trajectory`, and we dont need to set it in `trajectory-filter`.

cc. @pazeshun, @708yamaguchi


knorth55:
@pazeshun can i close this PR and merge it to #361 ?

pazeshun:
Yes, please

PR #364, opened 2018-08-18
pr2-interface: Make use of return value of :move-gripper method

furushchev:
Currently `:move-gripper` just returns `nil` in any case.
This PR is for updating the return value to make users know whether moving gripper ends successfully or not.
This implementation returns `T` if either the goal of desired distance between gripper tips is reached or stalled by obstacles (e.g. When an object is held.), otherwise returns `NIL`.
If `:wait` is `nil`, return value is `T` if the goal is accepted and active, `NIL` otherwise.
I'm not sure this is a good specification, so any feedback is welcome.

A minor but important bugfix is also included.
https://github.com/jsk-ros-pkg/jsk_pr2eus/compare/master...furushchev:move-gripper?expand=1#diff-7974ca62f7389ec56d54a9080a63f58fL272


k-okada:
```
Message:"collsion avoidance motion is too fast 0.022"
```

k-okada:
if we merge this, can we close https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/209 ?

furushchev:
@k-okada I would say no. What are proposed in those PRs are:

In #209,
- `:wait` key is added to `:start-grasp` method (currently we must wait for grasping by default)
- `:get-grasp-result` method is added to get gripper position without moving grippers (currently gripper position can only be accessible as a return value of `:start-grasp`)

In this pull request:
- `:move-gripper` returns `T` or `NIL` according to the result of grasping action

Both are not related functionally.

PR #369, opened 2018-08-24
[pr2eus/speak.l] Fixed #368. When sound_play::SoundRequest has a :volume slot, set volume value.

### LINKED ISSUE #368: play-sound method could not play sound in kinetic.
We can not play sound by play-sound method in kinetic,
because sound-play has a volume slot and its default is 0.0.
Related to #274 

In kinetic, released sound-play's version is 0.3.3.

https://github.com/ros-drivers/audio_common/blob/master/sound_play/CHANGELOG.rst

```
sudo dpkg --list | grep sound-play
ii  ros-kinetic-sound-play                                      0.3.3-0xenial-20180809-161433-0800                    amd64        sound_play provides a ROS node that translates commands on a ROS topic (robotsound) into sounds.
```

iory:
Fixed #368 

I added a volume key value.
This function is enabled for indigo and kinetic user.

### DIFF
--- a/pr2eus/test/speak-test.l
+++ b/pr2eus/test/speak-test.l
@@ -29,7 +29,7 @@
 
 (deftest test-speak-jp ()
   (assert (speak-jp "こんにちは"))
-;;  (assert (speak-jp "こんにちは" :wait t))
+  (assert (speak-jp "こんにちは" :wait t))
 )
 
 (deftest test-ri-speak-en ()
@@ -39,7 +39,7 @@
 
 (deftest test-ri-speak-jp ()
   (assert (send *ri* :speak-jp "やあ"))
-;;  (assert (send *ri* :speak-jp "やあ" :wait t))
+  (assert (send *ri* :speak-jp "やあ" :wait t))
 )
 
 (run-all-tests)


pr2eus/speak.l
```diff
@@ -24,6 +25,9 @@
       (t
        (send msg :sound sound_play::SoundRequest::*say*)
        (send msg :arg (string sound))))
+    (when (string> (unix::getenv "ROS_DISTRO") "indigo")
```
furushchev: This comparison is not straightforward and leads malfunction since we on real robots use the fixed version of the `sound_play` package.

furushchev: How about chekcing if the message has `:volume` accessor  or other methods (e.g. checking md5)?
(Sorry, I cannot check they work)

iory: I see. Using find-method is suitable in this case?
https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/369/commits/daa1cb583b8ad031f720121d57ab1534c1ad5293#diff-26b37c5ff1eb538e9769876c2116fcc8R28

iory: @furushchev 
Modified. Please re-review this PR .
https://github.com/jsk-ros-pkg/jsk_pr2eus/commit/daa1cb583b8ad031f720121d57ab1534c1ad5293#diff-26b37c5ff1eb538e9769876c2116fcc8R28

furushchev:
@iory Thanks for the fix.
- Did you check on both environments where the message has the `volume` parameter and w/o?
- Can this be included in tests?

k-okada:
@iory please confirm  https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/374  fails and your patch fix it.

pr2eus/speak.l
```diff
@@ -24,6 +25,9 @@
       (t
        (send msg :sound sound_play::SoundRequest::*say*)
        (send msg :arg (string sound))))
+    (when (find-method msg :volume)
```
k-okada: @iory can you add comment something like
"since sound_play version x.x.x, the message including volume slot. c.f. github.com/..../pulls/..."

708yamaguchi:
Now, fetch cannot speak bacause `volume` value is always 0.
This Pull Request will solve this fetch's problem.

pr2eus/speak.l
iory: OK. I add comment.

iory:
OK. I add comment and modified title of this PR.

k-okada:
The contents of this PR is merged by #379 with test code, this PR only contains enable `speak-jp` with `wait: t`

PR #363, opened 2018-08-14
Fix time_from_start in motion-planned angle-vector-sequence

pazeshun:
Include #372 

When I use `:angle-vector-motion-plan` in pr2eus_moveit with list of two angle-vectors, the last angle-vector is skipped.
This is because trajectory points for second angle-vector starts from `time_from_start: 0` again.
I add test to detect this bug and fix.
```
$ rostopic echo /robot/limb/left/follow_joint_trajectory/goal 
header: 
  seq: 1
  stamp: 
    secs: 1534242824
    nsecs: 729660623
  frame_id: ''
goal_id: 
  stamp: 
    secs: 0
    nsecs:         0
  id: 1534242820729731921_/default_robot_interface_1534239753379624356_80342_/robot/limb/left/follow_joint_trajectory_12
goal: 
  trajectory: 
    header: 
      seq: 1
      stamp: 
        secs: 1534242824
        nsecs: 729660623
      frame_id: ''
    joint_names: ['left_s0', 'left_s1', 'left_e0', 'left_e1', 'left_w0', 'left_w1', 'left_w2']
    points: 
      - 
        positions: [1.2676694342046062, 0.6743298063427532, -1.9387100136411437, 2.103555420218015, 1.6277475880211183, 0.6973072658200672, 0.3449746277554351]
        velocities: []
        accelerations: []
        effort: []
        time_from_start: 
          secs: 0
          nsecs:         0
      - 
        positions: [1.2172655913958197, 0.6572224481698226, -1.9436173999322994, 2.0632726139582402, 1.6479222227612684, 0.6763317184891946, 0.3307340925707729]
        velocities: []
        accelerations: []
        effort: []
        time_from_start: 
          secs: 0
          nsecs: 505630592
      - 
        positions: [1.1668617485870332, 0.6401150899968919, -1.9485247862234552, 2.022989807698467, 1.6680968575014194, 0.655356171158322, 0.31649355738611074]
        velocities: []
        accelerations: []
        effort: []
        time_from_start: 
          secs: 0
          nsecs: 786699648
      - 
        positions: [1.1164579057782467, 0.6230077318239613, -1.9534321725146109, 1.9827070014386932, 1.6882714922415705, 0.6343806238274494, 0.3022530222014488]
        velocities: []
        accelerations: []
        effort: []
        time_from_start: 
          secs: 1
          nsecs:   7248274
      - 
[...]

PR #362, opened 2018-08-11
debug travis

k-okada:
https://travis-ci.org/jsk-ros-pkg/jsk_pr2eus/jobs/414423684

```
--
Still running (32 of 60): catkin run_tests -i --no-deps --no-status pr2eus pr2eus_moveit pr2eus_tutorials -j1 -p1 --make-args -j1 --
Still running (33 of 60): catkin run_tests -i --no-deps --no-status pr2eus pr2eus_moveit pr2eus_tutorials -j1 -p1 --make-args -j1 --

The job exceeded the maximum time limit for jobs, and has been terminated.

```

And if we run locally, it fails as follows

```
root@60bff65c8efa:~# catkin_test_results --verbose
Full test results for 'build/pr2eus_moveit/test_results/pr2eus_moveit/rosunit-test_pr2eus_moveit.xml'
-------------------------------------------------
<testsuite tests="4" failures="1" disabled="0" errors="0" time="75" name="AllTests">
  <testcase name="all-test" status="run" time="0" classname="all-test">
  </testcase>
  <testcase name="test-angle-vector-motion-plan" status="run" time="5" classname="test-angle-vector-motion-plan">
  </testcase>
  <testcase name="test-angle-vector-sequence-motion-plan" status="run" time="53" classname="test-angle-vector-sequence-motion-plan">
  </testcase>
  <testcase name="test-move-end-coords-plan" status="run" time="8" classname="test-move-end-coords-plan">
   <failure message="collsion avoidance motion is too fast 0.0" type="AssertionError">
Test:(&gt; tm-diff 3)
Trace:";; ik error at (-31 . NO_IK_SOLUTION)
ds-plan]
motion-plan]
gle(11.4979) violate min-angle(11.5)
l
))
"
Message:"collsion avoidance motion is too fast 0.0"
   </failure>
  </testcase>
  <testcase name="test-moveit-fastest-trajectory" status="run" time="9" classname="test-moveit-fastest-trajectory">
  </testcase>
</testsuite>

-------------------------------------------------
build/pr2eus_moveit/test_results/pr2eus_moveit/rosunit-test_pr2eus_moveit.xml: 4 tests, 0 errors, 1 failures, 0 skipped
Summary: 59 tests, 0 errors, 1 failures, 0 skipped
```

PR #370, opened 2018-08-27
fix velocity time scaling

knorth55:
@pazeshun can you review this?

### DIFF
--- a/pr2eus_moveit/euslisp/robot-moveit.l
+++ b/pr2eus_moveit/euslisp/robot-moveit.l
@@ -643,18 +643,26 @@
             (time-step (/ 1 (float (1- size))))
             (time-scale (/ (/ total-time 1000.0) orig-total-time)))
        (dolist (pt points)
-         (when clear-velocities
-           (send pt :velocities #f())
-           (send pt :accelerations #f()))
-         (send pt :velocities
-               (map float-vector #'(lambda (x) (/ x time-scale)) (send pt :velocities)))
+         (if clear-velocities
+           (progn
+             (send pt :velocities #f())
+             (send pt :accelerations #f()))
+           (progn
+             (send pt :velocities
+                   (map float-vector #'(lambda (x) (/ x time-scale)) (send pt :velocities)))
+             (send pt :accelerations
+                   (map float-vector #'(lambda (x) (/ x (expt time-scale 2))) (send pt :accelerations)))))
          (send pt :time_from_start (ros::time (+ start-offset-time (* time-scale (send (send pt :time_from_start) :to-sec))))))
        (dolist (pt mdof-points)
-         (when clear-velocities
-           (send pt :velocities nil)
-           (send pt :accelerations nil))
-         (send pt :velocities
-               (map float-vector #'(lambda (x) (/ x time-scale)) (send pt :velocities)))
+         (if clear-velocities
+           (progn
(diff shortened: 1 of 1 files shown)


k-okada:
we'd like to create test code of this, there would be two ideas
1. run moveit and send `:angle-vector-motion-plan` and check result
2. redefine `(defmethod robot-interface
(:angle-vector-make-trajectory))` in test code and force returns the
trajectory for test.

If we can run moveit on travis very stably , 1 would be good, if not,
we can choose 2. the return value can be manually obtained by calling
`
`(:angle-vector-make-trajectory` and store the result S expression.

test could should be
1. at least, checks if the error is not occour
2. It would be better if we can visualize output `rqt_joint_trajectory_plto`
3. ofcourse, if we can check the output programmaly, that would be great




--
◉ Kei Okada

--
◉ Kei Okada



2018年8月27日(月) 18:23 Shingo Kitagawa <notifications@github.com>:

pazeshun:
@knorth55 I'm very sorry for forgetting this PR. I think this works, but I prefer scaling velocity only when `clear-velocities` is nil.
@k-okada 

> we can choose 2. the return value can be manually obtained by calling

This is partially tried in [this commit](https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/363/commits/b70fd97d2ba32035310bcb15162cfb7b9d2a6207) and works in local (= detect error).
I also tried to this on travis, but at that time, problem in https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/362#issue-207749700 prevents travis.
Should I send PR having only this test?

knorth55:
I updated accelerations time scale and other review.
However, I don't know which robot move with accelerations command, so I haven't tested yet.

k-okada:
> problem in #362 (comment) 

I think this has been released, please check again @knorth55

pr2eus_moveit/euslisp/robot-moveit.l
```diff
@@ -643,14 +643,26 @@
...
-         (when clear-velocities
-           (send pt :velocities nil)
-           (send pt :accelerations nil))
+         (if clear-velocities
+           (progn
+             (send pt :velocities nil)
+             (send pt :accelerations nil))
+           (progn
+             (send pt :velocities
+                   (map float-vector #'(lambda (x) (/ x time-scale)) (send pt :velocities)))
```
pazeshun: `(send pt :velocities)` should not be float-vector, but should be `geometry_msgs::Twist` because `pt`'s msg type is `trajectory_msgs::MultiDOFJointTrajectoryPoint`

pr2eus_moveit/euslisp/robot-moveit.l
```diff
@@ -643,14 +643,26 @@
...
+             (send pt :accelerations
+                   (map float-vector #'(lambda (x) (/ x (expt time-scale 2))) (send pt :accelerations)))))
```
pazeshun: `(send pt :accelerations)` should not be float-vector, but should be `geometry_msgs::Twist` because `pt`'s msg type is `trajectory_msgs::MultiDOFJointTrajectoryPoint`

pazeshun:
@knorth55 I created test in #376, which should become passed with this PR.
Please check that.

pazeshun:
I created new version of this as #377

knorth55:
move to #377

PR #372, opened 2018-09-07
Add test for concatenation in :angle-vector-make-trajectory 

pazeshun:
Included in #363 

Extract test code from #363 .
Travis of this PR will fail.


pazeshun:
```
<testcase name="test-avs-to-angle-vector-make-trajectory" status="run" time="0" classname="test-avs-to-angle-vector-make-trajectory">

   <failure message="time_from_start is not arranged in ascending order" type="AssertionError">
```
Test is correctly failed

PR #375, opened 2018-09-19
Add test for start-offset-time with avs 

pazeshun:
Included in #365 
Extracted test code from #365 
Travis of this PR will fail.

(from https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/365#issue-209352941)

## Minor but important fix
- Speed up tests to pass travis
  - Use `init-pose` in `test-angle-vector-sequence-motion-plan` because using `reset-manip-pose` moves torso a lot and very slow
  - Reduce `start-offset-time` in tests


pazeshun:
After https://github.com/jsk-ros-pkg/jsk_pr2eus/commit/1536883e54600cb5561896c683288f9442d24560, test's time becomes short.

test-avs-to-angle-vector-make-trajectory: 1 sec
test-angle-vector-motion-plan: 8 sec
test-angle-vector-sequence-motion-plan: 32 sec
test-move-end-coords-plan: 15 sec
test-moveit-fastest-trajectory: 16 sec
test-start-offset-time: 18 sec
test-start-offset-time-with-avs: 30 sec

And `test-start-offset-time-with-avs` becomes tested before timeout

pazeshun:
Test correctly failed in both indigo and kinetic.

PR #376, opened 2018-09-19
Add test for time scaling of velocity and acceleration

pazeshun:
Test for ~~#370~~#377
Travis of this PR will be failed.

`:trajectory-filter` in `robot-moveit.l` scales positions, but leaves velocities and accelerations as they are when `:clear-velocities nil` is passed.
This violates smoothness of trajectory when the robot follows velocities or accelerations.
This test catches that bug.
That bug will be fixed in ~~#370~~#377

As many robots in JSK don't use velocities and accelerations currently, that bug hasn't appear.
However, there are some robots requiring velocities (information from @knorth55 ).
So we should fix that bug for future.


pazeshun:
Test correctly failed in indigo, too.

PR #377, opened 2018-09-20
Fix velocity time scaling

pazeshun:
Updated version of #370 
- Resolve conflicts with origin/master
- Fix scaling of multi dof joint trajectory (https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/370#pullrequestreview-156665705)
- Rebased on #376 

This violates smoothness of trajectory when the robot follows velocities or accelerations.

> As many robots in JSK don't use velocities and accelerations currently, that bug hasn't appear.
However, there are some robots requiring velocities (information from @knorth55 ).
So we should fix that bug for future.

(from https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/376#issue-216546619)


pazeshun:
@knorth55 Could you test this with your robot?
@furushchev Could you test this with a robot with moving base?

pazeshun:
@iory Could you test this with HSR?

iory:
OK.　I'll test it.

iory:
I checked this PR with HSR and it correctly works!

PR #379, opened 2018-10-31
fix speak-test.test to fail on volume==0

k-okada:
add code to reproduce #368, whcih has newer sound_play message with volume and the default values is 0


iory:
When will this RP be merged?

k-okada:
Another idea is to generate dummy `text2wave.sh` https://gist.github.com/k-okada/47cb17d3648d8173230a89d85242e7e7, but `soundplay_node.py` still requires desktop (?) environment.

```
#!/bin/bash

cat > /tmp/text2wave <<'EOF'
#!/bin/bash
# create dummy /tmp/text2wave for docker environment

# for festival compaibility
ARGS=`echo $@ | sed 's/\-eval\ ()//' | sed 's/\-eval\ (\(.\+\))\ \(.\+\)$/\2\ \-p\ \1/'`
eval set -- "$ARGS"

INPUT_FILE=$1;shift;
OUTPUT_FILE=/tmp/_voice_text_$$.wav
while getopts p:o: OPT
do
  case ${OPT} in
    "o")
      OUTPUT_FILE=$(readlink -f "$OPTARG")
      ;;
  esac
done

if [ ! -e $INPUT_FILE ]; then
  echo "$INPUT_FILE not found" 1>&2
  exit 1
fi

sleep 1
echo -e "RIFFF\0\0\0WAVEfmt \x10\0\0\0\x01\0\x02\0\x80\xbb\0\0\0\xee\x02\0\x04\0\x10\0LIST\x1a\0\0\0INFOISFT\x0e\0\0\0Lavf46.40.101.data\0\0\0" > $OUTPUT_FILE

exit 0
EOF

chmod u+x /tmp/text2wave
```

PR #380, opened 2018-12-01
pr2-interface.l: Add optional key arguments for :angle-vector-with-constraint

furushchev:
- Add docstring for `:angle-vector-with-constraint` method
- add optional key arguments for the method
  - `:revert-if-fail`
  - `:div`
  - `:initial-angle-vector`


pazeshun:
I don't know much about `:angle-vector-with-constraint`, but is it similar to `:angle-vector` with `:end-coords-interpolation t`?
If so, please consider to use and improve end-coords-interpolation if you have time.

furushchev:
Partially yes, it's the similar but this is also often used for avoiding rotation of infinite rotational joints.
I agree to discuss future migration to the function you mentioned anyway.

PR #381, opened 2018-12-11
[pr2eus/robot-interface.l] Modified warning message in case of we can not connect to follow joint trajectory server

iory:
This PR modified the warning message in case of we can not connect to follow joint trajectory server,
because```goal``` topic is not always comming. 
So we should check the ```status``` topic.


pazeshun:
@iory kindly ping

iory:
I modified warn message
`make sure that you can run 'rostopic echo /~A/goal' and 'rostopic info /~A/goal'` to `make sure that you can run 'rostopic info /~A/goal'`
`

pazeshun:
Is warn message correct?
Current implementation may allow `rostopic info` even with kinematic simulator, because robot-interface subscribes it.
Please check the behaviour.

iory:
I modified warn message. I asked people where subscribers and publishers exist.

iory:
Thanks you, too!

iory:
@k-okada 
Could you please review modification of warning messages?

PR #382, opened 2018-12-12
fix typo and add pr2-init in pr2eus-moveit.l for PR2 + moveit

knorth55:


pazeshun:
1. So did https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus_moveit/test/test-pr2eus-moveit.l#L81 call new function in travis test of this PR?
2. Why did previous code work with wrong parenthesis close?

pr2eus_moveit/euslisp/pr2eus-moveit.l
```diff
@@ -1,4 +1,8 @@
+(require :robot-interface "package://pr2eus/robot-interface.l")
 (require :robot-moveit "package://pr2eus_moveit/euslisp/robot-moveit.l")
+(require :pr2 "package://pr2eus/pr2.l")
+(require :pr2-utils "package://pr2eus/pr2-utils.l")
```
pazeshun: I assume `:robot-interface`, `:pr2` and `:pr2-utils` are dependencies of `pr2-interface`.
Do we have to write them explicitly?
Writing them explicitly is not a common way, I think.
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus_openrave/pr2eus-openrave.l#L5

knorth55:
@k-okada Kindly ping. Can you merge this?

k-okada:
`pr2-init` is already defined in `pr2-interface.l` so please reuse them.
1. When we load `pr2-moveit.l`, rename original `pr2-init` as `pr2-init-orgp` and define newer version of `pr2-init` and call `pr2-init-org` within it.

k-okada:
Please refer to https://github.com/euslisp/EusLisp/blob/9cfd6550b551998bb9a29d78ef2cf28bba2ce59e/lisp/opengl/src/gluforeign.l#L75-L76 for rename original function.

PR #343, opened 2018-04-13
pr2eus: override clear-costmap namespace for pr2

furushchev:
Namespaces of services for clearing cost map or changing inflation range depends on configuration of `move_base` stack and different on each robots.
I added to override the existing method defined in `robot-interface.l` into `pr2-interface.l` for supporting pr2.

PR #396, opened 2019-10-01
Fix return value of clear-costmap function

YutoUchimi:
`clear-costmap` function and `:clear-costmap` method, whose namespace was fixed in #343, always return `t` even if a service call failed.
In this PR the function returns `t` after calling that service, and otherwise (if the service is not advertised yet) returns `nil`.

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -1744,8 +1744,11 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
    "Send signal to clear costmap for obstacle avoidance to move_base."
    (let ((srv-name (format nil "~A/clear_costmaps" (send move-base-action :name))))
      (if (ros::wait-for-service srv-name 0)
-         (call-empty-service srv-name)))
-   t)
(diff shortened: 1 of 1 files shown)


k-okada:
How about
```
(if (ros::wait-for-service srv-name 0)
          (call-empty-service srv-name)))))
```
?

YutoUchimi:
Thanks for your review!

- I wrote `nil` for return value explicitly, but it's ok even if the function implicitly returns `nil`.

- It seems `(call-empty-service srv-name)` returns empty service response, which doesn't have success/fail field.
  Which should we get, the service response or boolean value?
  ```
  $ rossrv show std_srvs/Empty 
  ---
  
  ```

YutoUchimi:
I searched under jsk-ros-pkg/* and start-jsk/*, and could not find any code which assumes EmptyResponse as a return value of `(call-empty-service)`.
So I changed  `(call-empty-service)` to return `t` in https://github.com/jsk-ros-pkg/jsk_roseus/pull/621

PR #400, opened 2019-10-04
made it possible to give a function like lambda to the target value of inverse kinematics

kosuke55:
made it possible to give a function like lambda to the target value of inverse kinematics.

For example, the following shake-cocktail.
````
(defun shake-cocktail ()
  (send *irtviewer* :title "shake-cocktail")
  (unless (boundp '*robot*)
    (setq *robot* (instance sample-robot :init)))
  (send *robot* :reset-pose)
  (send *robot* :newcoords (make-coords))

  (setq *obj* (make-cylinder 20 100))
  (send *obj* :set-color #f(1 1 0))
  (send *robot* :reset-pose)
  (objects (list *robot* *obj*))

  (send *robot* :inverse-kinematics
        (list (make-coords :pos #f(400 0 1000)))
        :move-target
        (list (send *robot* :larm :end-coords))
        :link-list
        (list (send *robot* :link-list
                    (send (send *robot* :larm :end-coords) :parent)
                    (car (send *robot* :larm :links))))
        :translation-axis (list t)
        :rotation-axis (list nil))

  (let* ((cnt 0.0))
    (do-until-key
     (incf cnt 0.1)
     (send *robot* :inverse-kinematics
           (list (make-coords :pos (float-vector (+ 400 (* 100 (sin cnt))) (* 50 (cos cnt)) 1000))
                 #'(lambda ()
                     (send (send (send *robot* :larm :end-coords) :copy-worldcoords)
                           :translate #f(0 0 100) :local)))
           :move-target
           (list (send *robot* :larm :end-coords)
                 (send *robot* :rarm :end-coords))
           :link-list
           (list (send *robot* :link-list
                       (send (send *robot* :larm :end-coords) :parent)
                       (car (send *robot* :larm :links)))
                 (send *robot* :link-list
                       (send (send *robot* :rarm :end-coords) :parent)
                       (car (send *robot* :rarm :links))))
           :translation-axis (list :z t)
           :rotation-axis (list nil :z))
     (send *obj* :newcoords (send (send *robot* :larm :end-coords) :copy-worldcoords))
     (send *irtviewer* :draw-objects))))
````

### DIFF
--- a/pr2eus/pr2-utils.l
+++ b/pr2eus/pr2-utils.l
@@ -42,18 +42,19 @@
 (defmethod pr2-robot
   (:select-target-arm
    (c)
-   (setq c (mapcar #'(lambda (x) (if (functionp x) (funcall x) x)) c))
    (cond ((and (consp c) (= 2 (length c))) ;; dual arm
+          (setq c (mapcar #'(lambda (x) (if (functionp x) (funcall x) x)) c))
 	  (let ((v0 (send self :inverse-transform-vector (send (car c) :worldpos)))
 		(v1 (send self :inverse-transform-vector (send (cadr c) :worldpos))))
 	    (if (> (elt v0 1) (elt v1 1)) (list :larm :rarm) (list :rarm :larm))))
          ((consp c) ;; single arm
+          (setq c (mapcar #'(lambda (x) (if (functionp x) (funcall x) x)) c))
 	  (let ((v (send self :inverse-transform-vector (send (car c) :worldpos))))
 	    (if (> (elt v 1) 0) :larm :rarm)) ;; check y-coords
           )
-	 (t         ;; single arm
-	  (let ((v (send self :inverse-transform-vector (send c :worldpos))))
-	    (if (> (elt v 1) 0) :larm :rarm)) ;; check y-coords
+         (t         ;; single arm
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by knorth55** [comment]
you pass `#'lambda` but you just `eval` (actually `funcall`) it just inside `:inverse-kinematics` method.
it seems that you can do the same thing as your example if you first calculate target coordinates before `:inverse-kinematics` and pass it.
why do you need this function?
do you need this function for more complicated use case?

**Comment by kosuke55** [comment]
I wanted to move both grippers while keeping their relative positions.
When the `shake-cocktail` sample function was executed with pr2, an error occurred in the `pr2-roboto` part, so I modified it.
But certainly I can do the same if I eval before giving it to ik.
Also, I don't do anything more complicated than this function, so should I close it?

**Comment by knorth55** [comment]
In your case, you don't need to modify this code because you just `funcall x` with no args.
```lisp
(setq c (mapcar #'(lambda (x) (if (functionp x) (funcall x) x)) c))
```
However, if you want to `funcall x` with args or slot values, it would be interesting.
But I have no idea about how to use that kind of complicated use case.

knorth55:
If you think you don't need this function and you check that you can do the same thing without this PR, you can close this.

PR #402, opened 2019-10-07
Updates to :stop-motion

Affonso-Gui:
The `:stop-motion` stops the current angle vector by sending the actual angle vector as a new command (and therefore overwriting the old command).

The problem is that is also does that when there is no ongoing action, and since it also have an `:wait-interpolation` it takes a few seconds to return from the call.

In this PR we:
1. Only overwrite commands when there are ongoing actions
1. Add a `wait` key argument to the `:stop-motion`, so the user can opt not to wait-interpolation.

@pazeshun

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -626,6 +626,7 @@ return t if interpolating"
              (setq cacts (gethash ctype controller-table)))
             (t
              (setq cacts controller-actions)))
+      (send self :spin-once)  ;; update interpolatingp
       (some #'(lambda (a) (send a :interpolatingp)) cacts)))
   (:wait-interpolation-smooth (time-to-finish &optional (ctype))
 "Return time-to-finish [msec] before the sent command is finished. Example code are:
@@ -667,9 +668,10 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
             (send-all controller-actions :time-to-finish)))))
   (:stop-motion (&key (stop-time 0))
    "Stop current executing motion"
-   (let ((av (send self :state :potentio-vector)))
-     (send self :angle-vector av stop-time)
-     (send self :wait-interpolation)))
+   (when (send self :interpolatingp)
+     (let ((av (send self :state :potentio-vector)))
+       (send self :angle-vector av stop-time)
+       (send self :wait-interpolation))))
   (:cancel-angle-vector
    (&key ((:controller-actions ca))
          ((:controller-type ct))
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -666,12 +666,14 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
               (send-all cacts :time-to-finish)))
            (t
             (send-all controller-actions :time-to-finish)))))
-  (:stop-motion (&key (stop-time 0))
+  (:stop-motion (&key (stop-time 0) (wait t))
    "Stop current executing motion"
    (when (send self :interpolatingp)
      (let ((av (send self :state :potentio-vector)))
        (send self :angle-vector av stop-time)
-       (send self :wait-interpolation))))
+       (if wait
+           (send self :wait-interpolation)
+           t))))
   (:cancel-angle-vector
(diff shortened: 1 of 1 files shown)


k-okada:
Do you need this because I did not merged https://github.com/start-jsk/rtmros_common/pull/765#issuecomment-392741195 ?

Affonso-Gui:
@k-okada This is not directly related to https://github.com/start-jsk/rtmros_common/pull/765#issuecomment-392741195.

I noticed this problem when adding a call of `:stop-motion` to the default sigint behaviour, which led me to waiting a couple seconds every time C-c is pressed, even though there was no ongoing motion.

YoheiKakiuchi:
Why this ```it also have an :wait-interpolation it takes a few seconds to return from the call.``` happens.

I saw the code, by this ```(send self :angle-vector av stop-time)```,
the duration of this command was limited by min-time (= 1.0 second).
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/robot-interface.l#L374

k-okada:
so why not using :cancel-angle-vector
--
◉ Kei Okada



2019年10月8日(火) 14:48 Affonso, Guilherme <notifications@github.com>:

Affonso-Gui:
@YoheiKakiuchi Sorry, saying that it takes a few seconds was a bit exaggerated. It seems to take exactly 1 second
```lisp
$ (bench (send *ri* :stop-motion))
[ INFO] [1570526342.652453956]: wait-interpolation debug: start
[ INFO] [1570526343.678601001]: wait-interpolation debug: end
;; time -> 1.08321[s]
(nil nil nil nil)
```

@k-okada `:cancel-angle-vector` leads to an abrupt stop, while `:stop-motion` makes the robot stops smoothly. I thought that it would be best fitted for a demo to use the `:stop-motion` in planned interruptions.

k-okada:
i see, please add that to doc string.
--
◉ Kei Okada



2019年10月8日(火) 18:25 Affonso, Guilherme <notifications@github.com>:

Affonso-Gui:

pr2eus/robot-interface.l
```diff
@@ -665,16 +666,19 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
               (send-all cacts :time-to-finish)))
            (t
             (send-all controller-actions :time-to-finish)))))
-  (:stop-motion (&key (stop-time 0))
-   "Stop current executing motion"
-   (let ((av (send self :state :potentio-vector)))
-     (send self :angle-vector av stop-time)
-     (send self :wait-interpolation)))
+  (:stop-motion (&key (stop-time 0) (wait t))
+   "Smoothly stops the motion being executed by sending the current joint state to the joint trajectory action goal"
```
k-okada: @Affonso-Gui how about  adding “, uses are expected to use :cancel-angle-vector, for most cases” or something.

Affonso-Gui: Are uses expected to use :cancel-angle-vector for most cases?

k-okada: if it is supported, what is the reason to use :stop-motion ?

Affonso-Gui: Both `:cancel-angle-vector` and `:stop-motion` are (should be) supported on all robots that use follow-joint-trajectory controllers.

> :cancel-angle-vector leads to an abrupt stop, while :stop-motion makes the robot stops smoothly. 

Although there are robots in which `:cancel-angle-vector` is smooth enough, prioritizing speed or smoothness is task/motion dependent, so I don't believe we can generalize into using one over the other.

k-okada: I see, then how about adding "note that :cancel-angle-vector leads to an abrupt stop, while :stop-motion makes the robot stops smoothly.", ":stop-time is transition time for stopping motion, but the actual interpolation duration is limited by :min-time of :angle-vector, which is set to 1.0 [sec] by default" etc, to avoid confusion like my case.

Affonso-Gui: @k-okada Added more information to the docstring. Please check to see if it's clear enough.

PR #412, opened 2019-11-04
use map frame for :state :worldcoords

knorth55:
`pr2-interface.l` uses `world` coords for `:state :worldcoords` (instead `robot-interface` uses `map` coords for `:state :worldcoords`)
I update to use `map` for `:state :worldcoords`


k-okada:
I think this is for elevator move, please check again

knorth55:
I see. So, we should use `world` for `:state :worldcoords`.

PR #418, opened 2019-11-08
[pr2eus] Add correction arg in :move-to-wait

knorth55:
add `:correction` arg in `:move-to-wait`
if you set `:correction nil`, the correction step will be skipped.
related #406


taichiH:
How about add correction argument in `:move-to` function?
Should I PR to your branch?

knorth55:
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/ba01746e9494a94fe82ed82b38d190cb60594de4/pr2eus/robot-interface.l#L1336-L1340
all args is passed to `:move-to-wait`, so you can pass `:correction` arg in `:move-to` function as follows;
`(send *ri* :move-to coords :correction nil)`

taichiH:
あ、すいません。勘違いしてました。
ありがとうございます。

PR #419, opened 2019-11-14
[pr2eus_moveit] fix cylinder collision object msg

knorth55:
the first element of dimensions is height, and the second is radius.
I fixed its order.

see here
http://docs.ros.org/melodic/api/shape_msgs/html/msg/SolidPrimitive.html


knorth55:
@k-okada Travis test passed. Could you merge this PR?

PR #421, opened 2019-11-15
[pr2eus_moveit] add method to delete collision objects by id

knorth55:
I add `:delete-object-by-id` and `delete-attached-object-by-id` methods.
We can delete collision objects by id (such as `ground`, `keepout`) with these methods.


knorth55:
@k-okada Travis test passed. Could you review and merge this PR?

PR #422, opened 2019-11-19
Fix bug on joint-move-over-180 with tm :fast

Affonso-Gui:
On pr2, calls to `:angle-vector` with time as `:fast` lead to `number expected` errors when `:check-continuous-joint-move-over-180` is true.

This PR avoids this by setting all the sub-times as `:fast` in this case.


Affonso-Gui:

Affonso-Gui:
@knorth55

k-okada:
Can we add test code to catch this?

Affonso-Gui:
@k-okada Added.

Affonso-Gui:
@k-okada Tests pass/fail as expected. Can we merge this?

PR #431, opened 2020-04-12
[pr2eus_moveit] Add test for collision-object-publisher

708yamaguchi:
I added test for `collision-object-publisher`

In the test, PR2 moves left arm between blocks.
PQP collision check is executed during the interpolation.

Before move
before_plan_side

before_plan

After move
after_plan_side

after_plan


k-okada:
thaks add test is always welcome;
not sure why but test is actuall passing https://travis-ci.org/github/jsk-ros-pkg/jsk_pr2eus/builds/674023155

PR #438, opened 2020-05-14
[pr2eus_moveit] Add timeout arg for computing IK and increase timeout for test

708yamaguchi:
In this pull request, I added `timeout` arg to some functions of `robot-moveit.l` and set the `timeout` in the travis test longer.

This pull request aims to prevent the travis test failure of `pr2eus_moveit`, such as:
https://travis-ci.org/github/jsk-ros-pkg/jsk_pr2eus/builds/674758416
https://travis-ci.org/github/jsk-ros-pkg/jsk_pr2eus/builds/686633237

These tests fail here:
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/867460fbf023dd004b0f59cdf1123c62c8adee1f/pr2eus_moveit/test/test-pr2eus-moveit.l#L208-L214

Error messages are as follows:
```
  <testcase name="test-move-end-coords-plan" status="run" time="15" classname="test-move-end-coords-plan">
   <failure message="collsion avoidance motion is too fast 0.001" type="AssertionError">
Test:(&gt; tm-diff 3)
Trace:";; ik error at (-31 . NO_IK_SOLUTION)
```

In detail, I think the reason is that
1.  IK calculation in `:move-end-coords-plan` fails
2. PR2 moves the arm little by `:move-end-coords-plan`
3. `tm-diff` becomes too short.

The reason why the IK calculation sometimes fails may be that its `timeout` is short.
In my environment, the IK calculation sometimes takes ~0.1 [s], but the default `timeout` is 0.05 [s] in `robot-moveit.l`.

I checked the IK calculation execution time by this code.
https://github.com/ros-planning/moveit/blob/50e3c21081c03a106ac1157d8a8687710bcc0398/moveit_kinematics/kdl_kinematics_plugin/src/kdl_kinematics_plugin.cpp#L409-L410

The default `timeout` in `robot-moveit.l`:
https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/867460fbf023dd004b0f59cdf1123c62c8adee1f/pr2eus_moveit/euslisp/robot-moveit.l#L173

I am not confident, but I believe this pull requests improve success rate of the travis test.

By the way, I also tried to use `attempts` field in `moveit_msgs/PositionIKRequest`, but this field seems to be unavailable now.
https://github.com/ros-planning/moveit/pull/1288


knorth55:
:+1: for this PR.
good work!

PR #439, opened 2020-06-28
[pr2eus] fix to support namespace

Naoki-Hiraoka:
This PR enables to initialize `pr2-interface` under namespace.
```
(setq *ri* (instance pr2-interface :init :namespace "PR2")) 
```

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -1274,14 +1274,14 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
      (when base-controller-action-name
        (setq move-base-trajectory-action
              (instance ros::simple-action-client :init
-                       base-controller-action-name
+                       (if namespace (format nil "~A~A" namespace base-controller-action-name) base-controller-action-name)
                        control_msgs::FollowJointTrajectoryAction
                        :groupname groupname))
        (unless (send move-base-trajectory-action :wait-for-server 3)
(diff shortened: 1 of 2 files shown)


k-okada:
@Naoki-Hiraoka nice catch!
please check https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/440. It is alwyas better to have simple test code before you add PR.

pr2eus/robot-interface.l
```diff
@@ -1274,14 +1274,14 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
      (when base-controller-action-name
        (setq move-base-trajectory-action
              (instance ros::simple-action-client :init
-                       base-controller-action-name
+                       (if namespace (format nil "~A~A" namespace base-controller-action-name) base-controller-action-name)
```
k-okada, line 1277: @Naoki-Hiraoka Are you sure about `"~A~A"` ? This assume user set `:base-controller-action-name "/base_controller/follow_joint_trajectory`, but there might be a change to set  `:base-controller-action-name "base_controller/follow_joint_trajectory`.

I think `/namespace//base_controller/follow_joint_trajectory` does not look nice, but it won't do any harm.
```
2.irteusgl$ (ros::subscribe "/hoge//hoge" std_msgs::String #'identity)
```
```
$ rostopic list
/hoge/hoge
```

k-okada:
test code was added in #440

PR #441, opened 2020-08-06
[pr2eus] add provide in pr2-interface.l and pr2-utils.l

knorth55:
add `(provide :pr2-interface "pr2-interface.l")`


Affonso-Gui:
Is there something you really don't want to be loaded twice or is this just for code unification?
If it's the latter, could you also add it on `pr2-utils.l`?

knorth55:
the former case, I mean.
I just want to load this file once, but it doesn't accelerate loading files in my program that much.

Affonso-Gui:
Not that I'm opposing this PR (I think it is good practice to add `provides` in files that are usually `required`), but I don't believe that loading pr2-interface multiple times is slowing your program down.

All of the files loaded from it are `required`, and `ros::load-ros-manifest` also have double-loading guards, so even if you load it twice not much harm is done.

```lisp
irteusgl> (bench (load "package://pr2eus/pr2-interface.l"))
;; time -> 2.54901[s]
t
irteusgl> (bench (load "package://pr2eus/pr2-interface.l"))
;; time -> 0.019368[s]
t
```

knorth55:
yes, you are right.
I don't see much improvement for loading pr2-interface.l multiple times.
i make this pr to make the load faster but it doesn't affect much.
but now i think it is better to have provide for formatting.

--
Shingo Kitagawa
the University of Tokyo, JSK Lab
s-kitagawa@jsk.imi.i.u-tokyo.ac.jp


2020年8月6日(木) 15:10 Affonso, Guilherme <notifications@github.com>:

Affonso-Gui:
So... adding in `pr2-utils.l` as well?

knorth55:
ok i will

knorth55:
i updated and added pr2-utils.l

Affonso-Gui (approved):

PR #445, opened 2020-10-11
[pr2eus] Fix doc about timeout of :wait-interpolation

pazeshun:
While creating #444 , I found that the argument `timeout` of `:wait-interpolation` means a timeout period of each controller, not the whole `:wait-interpolation`.
This PR specifies it in the docstring of `:wait-interpolation`.

cc. @Naoki-Hiraoka @Kanazawanaoaki


k-okada (approved):
improving doc is always welcome

PR #447, opened 2020-10-29
add comments to :move-to function

k-okada:


pr2eus/robot-interface.l
```diff
@@ -1354,6 +1354,9 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
   ;;
   (:move-to
    (coords &rest args &key (no-wait nil) &allow-other-keys)
+   "move the robot to coordinates in reference defined in :frame-id keyward. The default frame is \"world\"
+    Return t if th if the robot reached the goal. By defaut, it blocks until the robot reached the goal.
```
knorth55, line 1358:
```suggestion
    Return t if the robot reached the goal. By defaut, it blocks until the robot reached the goal.
```

knorth55 (changes requested):

pr2eus/robot-interface.l
708yamaguchi, line 1358: By defaut
->
By default

708yamaguchi (changes requested):

pr2eus/robot-interface.l
```diff
@@ -1354,6 +1354,9 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
   ;;
   (:move-to
    (coords &rest args &key (no-wait nil) &allow-other-keys)
+   "move the robot to coordinates in reference defined in :frame-id keyward. The default frame is \"world\"
```
mqcmd196, line 1357:
```suggestion
   "move the robot to coordinates in reference defined in :frame-id keyword. The default frame is \"world\"
```

k-okada:
thank you, fixed comment . Should we add documen tot `:move-to-wait`? I am sure that `:move-to-send`is private method, so we do not need documentation. How about `:move-to-wait` ? Is it also private method?

708yamaguchi:
I do not use `:move-to-wait` explicitly in a program, so in my opinion, `:move-to-wait` is a private method.

PR #456, opened 2021-01-06
Add minjerk-interpolation to :angle-vector and :angle-vector-sequence

708yamaguchi:
I add
- :minjerk-interpolation option to :angle-vector and :angle-vector-sequence
- test for :minjerk-interpolation

Related to https://github.com/euslisp/jskeus/pull/595


k-okada (approved):
LGM

PR #458, opened 2021-04-29
update jsk_travis to 0.5.21

knorth55:
update jsk_travis to 0.5.21


knorth55:
I update this PR to 0.5.21

knorth55:
travis passed! but travis-ci.org is still set as expected test.
@k-okada can you change the expected test?

PR #461, opened 2021-06-18
[pr2eus_moveit] angle-vector-motion plan runs normal angle-vector in kinematics simulator

knorth55:
this PR avoid error when we run `angle-vector-motion-plan` in `kinematics simulator`.
the way is just run `angle-vector` or `angle-vector-sequence` when `kinematics simulator` is on.
this PR is useful when we share the same code for 1) `real robot + moveit` and 2) `kinematics simulator`.
now we need to switch `angle-vector-motion-plan` and `angle-vector` for testing same code in 1) and 2).

### DIFF
--- a/pr2eus/robot-interface.l
+++ b/pr2eus/robot-interface.l
@@ -374,7 +374,7 @@
    (let* ((prev-av (send robot :angle-vector)))
      (send-all (gethash ctype controller-table) :push-angle-vector-simulation av tm prev-av)))
   (:angle-vector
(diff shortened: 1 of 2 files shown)


k-okada (changes requested):
@knorth55 could you provide simplest example code which reproduce the problem?

knorth55:
```
(load "package://pr2eus/pr2-interface.l")
(load "package://pr2eus_moveit/euslisp/pr2eus-moveit.l")

(pr2-init)
(send *rI* :angle-vector-motion-plan (send *pr2* :reset-manip-pose))
```

PR #464, opened 2021-07-31
[pr2eus_tutorials] Fix publish status

shmpwk:
During pr2eus_tutorials, robot status is showed on rviz. 
Previously, it only shows "click tabletop object in image_view2! though we can click both on rviz and on image_view2.
So I've changed the showing status 
from
Screenshot from 2021-07-31 21-30-36
to
Screenshot from 2021-07-31 21-31-13


708yamaguchi (approved):

pr2eus_tutorials/euslisp/pr2-tabletop-object-grasp.l
```diff
@@ -210,15 +210,15 @@
   (setq *grasp-state* "returning")
   (publish-grasp-state)
   (pr2-pregrasp-pose)
-  (publish-info "click tabletop object in Rviz!")
-  (publish-info "click tabletop object in image_view2!")
+  (ros::ros-info "click tabletop object in Rviz or in image_view2!")
```
k-okada, line 213:
```
(defun publish-info (&rest args)
  (let ((s (format nil (car args) (cdr args))))
    (ros::ros-info s)
    (publish-status 'info s)))
```
`publish-info` calls both `ros::ros-info` and 'publish-status`, so why you need to call `ros::ros-info` twice?

pr2eus_tutorials/euslisp/pr2-tabletop-object-grasp.l
```diff
@@ -210,15 +210,15 @@
...
+  (publish-status 'info (concatenate string "click tabletop object" '(#\newline) "in Rviz or in image_view2!"))
```
k-okada, line 214: may be `(format nil "click tabletop object~%in Rviz or in image_view2!")` seems simpler

pr2eus_tutorials/euslisp/pr2-tabletop-object-grasp.l
shmpwk: You're right. (I forgot why I don't use `publish-info`)

pr2eus_tutorials/euslisp/pr2-tabletop-object-grasp.l
shmpwk: Thank you for your advice, I improve it.

pr2eus_tutorials/euslisp/pr2-tabletop-object-grasp.l
shmpwk: This is because 
```
(ros::ros-info  (format nil "click tabletop object~%in Rviz or in image_view2!"))
```
called by 
```
(publish-info (format nil "click tabletop object~%in Rviz or in image_view2!"))
```
splits the sentence on the command line, which is slightly difficult to read.
Screenshot from 2021-11-17 17-14-10

PR #465, opened 2021-09-28
Update current-time slot in controller-action-client on real robot

Affonso-Gui:
The `controller-action-client class` has a `current-time` slot for keeping track of the `time_from_start` field in the action feedback message.
This field is only used in simulation to calculate the `time-to-finish`, but it shouldn't hurt to set it when on the real robot as well, specially since we are already calculating it.

I've found this information useful for my work on interruptions, but nevertheless this change should be discussed and tested (is the test suite working for all robots?)

@pazeshun


k-okada (commented):
I think this patch changes the simulation behavior so how about add slot variable like  `current-time-simulation` and apply your change on `:action-feedback-cb` ? 
I think current PR 
``
``

PR #426, opened 2019-12-15
Fix wide_stereo camera model parameters

k-okada:
rewrited version #403, because it just replace pr1012 settings with pr1040

--

update pr10*_urdf.yaml for pretty format, for make it easy to compare between robot and versions

use python command to convert from robot_description string format to xml format, use lxml because of pretty function https://www.ronrothman.com/public/leftbraned/xml-dom-minidom-toprettyxml-and-silly-whitespace/, instead of minidom or ElementTree

```
python -c "import yaml; import datetime; from lxml import etree; comment=etree.Comment(datetime.datetime.today().strftime(' generated from pr1040 on %Y/%m/%d ')); urdf=etree.XML(yaml.load(open('pr1040_urdf.yaml').read())); print(etree.tostring(comment, pretty_print=True, xml_declaration=True)); print(etree.tostring(urdf, pretty_print=True, xml_declaration=False));"
```

PR #466, opened 2021-11-17
add test code for (pr2) and (instance pr2-robot :init) (instance r2-sensor-robot :init)

k-okada:
follow ups of #426

as https://github.com/jsk-ros-pkg/jsk_robot/issues/1407#issuecomment-971313512 reported, it breaks existing API


Kanazawanaoaki (approved):
LGTM. It is backward compatible.

PR #472, opened 2022-01-09
[pr2eus_tutorials] fix look-at-hand

shmpwk:
In the pr2 grasping tutorials, the usage of `look-at-hand` is wrong.

Here is the simulation outcome that PR2 looks at rarm when grasping.
Screenshot from 2022-01-09 14-00-13

Here is the real outcome that PR2 looks at rarm when grasping.
Screenshot from 2022-01-09 13-57-39

### DIFF
--- a/pr2eus_tutorials/euslisp/pr2-tabletop-object-grasp.l
+++ b/pr2eus_tutorials/euslisp/pr2-tabletop-object-grasp.l
@@ -155,7 +155,7 @@
   (publish-grasp-state)
   (publish-info "pre grasp pose...")
   (send *ri* :stop-grasp *arm* :wait t)
-  (send *pr2* :head :look-at-hand *arm*)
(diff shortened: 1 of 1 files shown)


knorth55:
how about fixing the `look-at-hand` method in `pr2eus/robot-interface.l`?

708yamaguchi:
The correct usage is 

```
(send *pr2* :look-at-hand *arm*)
```
?


I think the following code works well.
```
(load "package://pr2eus/pr2-utils.l")

(pr2)
(objects (list *pr2*))

;; (send *pr2* :head :look-at-hand :rarm)
(send *pr2* :look-at-hand :rarm)
```

shmpwk:
@knorth55 @708yamaguchi 
I overlooked it, thank you for your advise.

k-okada (approved):
nice catch!

PR #475, opened 2022-04-05
Enable either-or tuckarm-pose

nakane11:
I made the initial pose for human to hold robot hand.
One arm make tuckarm-pose and the other move to ik-target position.

key :
- tuckarm --- select the arm tucked
- target-height --- increase ik-target height

Later I will reflect hand detection on target-height.

give-hand-pose


pr2eus/pr2-interface.l
```diff
@@ -738,6 +738,48 @@ Example: (send self :gripper :rarm :position) => 0.00"
...
+  (let* ((current-arm (check-tuckarm-pose :thre 40)) ;; nil rarm larm
+         (msec 2000))
+    
+      (when (or (eq current-arm tuckarm) (not current-arm))
+	(progn
+	  (send *pr2* :larm :angle-vector #f( 25 0 0 -121 0 -6 0))
+	  (send *pr2* :rarm :angle-vector #f(-25 0 0 -121 0 -6 0))
+	  (send *ri* :angle-vector (send *pr2* :angle-vector) msec)
+	  (send *ri* :wait-interpolation)
+	  ))
```
tkmtnt7000, line 751: インデントがタブ文字になっているのでスペースにするほうが好ましいと思います．

tkmtnt7000:
FYI:
emacsの設定ファイル(`.emacs.d/init.el`)にこんな感じで追加しておくとタブを打ったときにタブ文字が入らなくなって便利です．
```lisp
;;; tabにスペース4つを利用                                                                    
(setq-default tab-width 4 indent-tabs-mode nil)
```

Affonso-Gui:
1. `give-hand-pose` is demo-specific, I think we could leave it out of this PR
2. maybe we can re-use the `one-tuckarm-pose` in the original `pr2-tuckarm-pose` to avoid code repetition? Something like:
```lisp
(defun pr2-tuckarm-pose (&rest args)
  (one-tuckarm-pose)
  ...)
```
I can help you with that if you need.

pr2eus/pr2-interface.l
nakane11: Thank you for your advice. I updated emacs settings and removed tab characters. 36a70bd5977ab40ef88716d2acd66b606c31abb7

nakane11:
I edited  `pr2-tuckarm-pose` to select one arm or  both arms.
Since the original `one-tuckarm-pose` includes lift up motion, to call it twice seems to be inefficient. 

In updated program, one-tuckarm-pose is executed as below:
```
(pr2-tuckarm-pose :larm :inside :one)
```
Third argument decides if both arms are tucked. Any words except `:both` will work.

nakane11:
And I renamed local variable **side** in `pr2-tuckarm-pose` by 2fb4fae. It decides manual direction of tuckarm. 
To assign side to dir in `get-tuckarm` is ambiguous and confusing with new arg arm-side, .

Affonso-Gui (approved):

PR #476, opened 2022-04-06
Allow to set :use-torso in :move-end-pos

Affonso-Gui:
Currently `use-torso` is only referenced if `link-list` is null: https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/master/pr2eus/pr2-utils.l#L88-L104
This means that `:use-torso t` won't work when indirectly calling `:inverse kinematics` from other methods, such as
```lisp
(send *pr2* :rarm :move-end-pos ...)
(send *pr2* :rarm :inverse-kinematics ...)
```

In this PR we force-add the torso link to a given link-list if the user explicitly sets `:use-torso`. When link-list is non-nil use-torso defaults to nil for backward compatibility.
```lisp
(send *pr2* :rarm :move-end-pos #f(0 0 200) :world)  ;; torso is NOT used
(send *pr2* :rarm :move-end-pos #f(0 0 200) :world :use-torso t)  ;; torso is used
```


Affonso-Gui:
@k-okada Kindly ping.

PR #481, opened 2022-05-21
[pr2eus] Add key topic-name to :send-cmd-vel-raw

nakane11:
Currently `:send-cmd-vel-raw` publishes topic set in `robot-move-base-interface :init`. 
This PR enables specifying topic name when calling the method. It is useful when you want to use subclass such as `pr2-interface`.

Added key `topic-name` follows `:send-cmd-vel-raw` in [fetch-interface](https://github.com/jsk-ros-pkg/jsk_robot/blob/c91004358e0e60bdeca5d05f1aa96e01bf970947/jsk_fetch_robot/fetcheus/fetch-interface.l#L355-L364).


tkmtnt7000:
Nice PR!!

pr2eus/robot-interface.l
```diff
@@ -1592,17 +1592,23 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
...
    (when (send self :simulation-modep)
      (return-from :send-cmd-vel-raw t))
-   (unless (ros::get-topic-publisher (if namespace (format nil "~A/~A" namespace cmd-vel-topic) cmd-vel-topic))
-     (ros::advertise (if namespace (format nil "~A/~A" namespace cmd-vel-topic) cmd-vel-topic) geometry_msgs::Twist 1)
+   (setq vel-topic
+         (if topic-name
+             (if namespace
+                 (if (eq (char "/" 0) (char topic-name 0)) topic-name (format nil "~A/~A" namespace topic-name))
+                 topic-name)
+             (if namespace (format nil "~A/~A" namespace cmd-vel-topic) cmd-vel-topic)))
```
tkmtnt7000, line 1603: It may be better to use `let` instead of `setq` to set `vel-topic` as a local variable.

nakane11: Thanks for the review. I have revised it.

Affonso-Gui:
> It is useful when you want to use subclass such as pr2-interface.

What do you mean? Why don't you just set the correct topic name in the :init function? Or do we have multiple base_controller servers?

nakane11:
I just missed [args](https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/e77d628d85453bc32917152c7ae8a8ef0c9d0b63/pr2eus/pr2-interface.l#L46) in send-super*. I can set the topic in the :init function as you said.

I would like to use :send-cmd-vel-raw with (pr2-init), but to make pr2-interface with topic name is substitutable for it.

pr2eus/robot-interface.l
```diff
@@ -1592,17 +1592,23 @@ Return value is a list of interpolatingp for all controllers, so (null (some #'i
...
    (when (send self :simulation-modep)
      (return-from :send-cmd-vel-raw t))
-   (unless (ros::get-topic-publisher (if namespace (format nil "~A/~A" namespace cmd-vel-topic) cmd-vel-topic))
-     (ros::advertise (if namespace (format nil "~A/~A" namespace cmd-vel-topic) cmd-vel-topic) geometry_msgs::Twist 1)
+   (let ((vel-topic
+         (if topic-name
+             (if namespace
+                 (if (eq (char "/" 0) (char topic-name 0)) topic-name (format nil "~A/~A" namespace topic-name))
+                 topic-name)
+             (if namespace (format nil "~A/~A" namespace cmd-vel-topic) cmd-vel-topic))))
```
Affonso-Gui, line 1603:
```lisp
  (:send-cmd-vel-raw
   (x y d &key (topic-name cmd-vel-topic) ((:namespace ns) namespace)) ;; [m] [m] [degree]
    (if ns (setq topic-name (format nil "~A/~A" ns topic-name)))
    ;; ...
```

Affonso-Gui (commented):
Another coding suggestion, in case you are fine on setting `:namespace nil` for global topics (the first letter is not checked)

pr2eus/robot-interface.l
nakane11: Thank you for the helpful advice. It's simple and easy to understand.

Affonso-Gui (approved):

tkmtnt7000 (approved):

PR #484, opened 2022-08-27
.github/workflows/config.yml: enable noetic USE_DEB=true

k-okada:


k-okada:
https://github.com/PR2/pr2_common/pull/291 might be one reason for notiec regression, but it does not solve everything.
so we disable imu in robot_pose_ekf.

PR #485, opened 2022-10-12
Install roseus_resume interruption handler for pr2

knorth55:
this PR install `roseus_resume` interruption handler for `pr2-interface`.
this PR doesn't load `roseus_resume` when `roseus_resume` is not found.

cc. @Affonso-Gui

### DIFF
--- a/pr2eus/pr2-interface.l
+++ b/pr2eus/pr2-interface.l
@@ -82,6 +82,15 @@
        (setq joint-action-enable nil)
        (ros::ros-warn "~A is not respond, pr2-interface is disabled" action)
        (return)))
+   ;; install interruption handler
+   (warning-message 3 "Installing interruption handler...~%")
+   (roseus_resume:install-interruption-handler self
+       r-gripper-action
+       l-gripper-action
+       move-base-action
+       move-base-trajectory-action)
+   (roseus_resume:install-default-intervention self)
+   (warning-message 3 "...done.~%")
    t)
   ;;
   (:state (&rest args) ;; overwrite for jsk_maps/pr2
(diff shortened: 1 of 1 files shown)


pr2eus/pr2-interface.l
```diff
@@ -5,6 +5,17 @@
 (require :pr2-utils "package://pr2eus/pr2-utils.l")
 (require :robot-interface "package://pr2eus/robot-interface.l")
 (require :speak "package://pr2eus/speak.l")
+(defvar *enable-roseus-resume* (ros::rospack-find "roseus_resume"))
```
k-okada, line 8: should we put this code in robot-interface?

pr2eus/pr2-interface.l
```diff
@@ -81,6 +92,17 @@
        (setq joint-action-enable nil)
        (ros::ros-warn "~A is not respond, pr2-interface is disabled" action)
        (return)))
+   ;; install interruption handler
+   (if *enable-roseus-resume*
+     (progn
+       (warning-message 3 "Installing interruption handler...~%")
+       (roseus_resume::install-interruption-handler self
```
k-okada, line 99: I am not familer with roseus_resume API, but define
```
(:install-roseus-resume-interruption-handler (args)
 (if *enable-roseus-resume* (progn ...) (progn ....)
```
 in  robot-interface or robot-move-base-interface and add 
```
(send self :install-roseus-resume-interaption-handler (list r-gripper-action l-gripper-action))
```
in pr2-interface.l

- I am qurious why we do not have to specity joint-trajectory-action?
- if we need to install all actionclient, may be we try to find them from program code

Affonso-Gui: I think that adding a `:install-roseus-resume-interruption-handler` method in the robot interface with internal checking is a valid approach for portability.

> I am curious why we do not have to specify joint-trajectory-action?

We already have access to joint trajectory actions via the `controller-actions` slot in the robot interface. So we only need to pass the "additional" actions in initialization.
https://github.com/Affonso-Gui/roseus_resume/blob/eus10/euslisp/interruption-handler.l#L51-L58

> if we need to install all actionclient, may be we try to find them from program code

It is theoretically possible to loop through all slots/bindings/symbols and register all instances of `ros::simple-action-client`.

knorth55: can we add additional controllers into `controller-actions`?
if not, we can make a new slot for additional controllers such as `additional-controller-actions`.
If we can do this, we no more need to change codes in `fetcheus` or `baxtereus`, we only need to change `pr2eus/robot-interface.l`.

knorth55: also, we need to change the order.
`(send-super* :init args)` is normally called at the first of `:init`, and we need to move `:add-controller` before `send-super`.

Affonso-Gui:
@k-okada @knorth55  https://github.com/jsk-ros-pkg/jsk_pr2eus/pull/488

k-okada:
closed via #488

PR #492, opened 2022-11-25
[pr2eus] add simulation-speed-scale for speeding up kinematics simulation

knorth55:
this PR adds `simulation-speed-scale` args for `kinematics simulation` speed up.
now all the motion and the movements are running with fixed speed, but we can adjust the speed by this argument.

### x10 speed up

https://user-images.githubusercontent.com/9300063/203938555-e6e94918-8cdd-4b43-816d-06c5a2c5006a.mp4


tkmtnt7000:
I usually use this in my local development of robot demostrations and this feature is helpful to reduce debugging time.
Time is money.

PR #495, opened 2023-06-07
[pr2eus] Add header timestamp to actiongoal topic in play-sound

tkmtnt7000:
Now, when we execute something like `(send *ri* :speak-jp "こんにちは")`, sound action goal topic has no headertimestamp.
I fixed it in this PR.
```
# before this PR
$ rostopic echo /robotsound_jp/goal
header: 
  seq: 0
  stamp: 
    secs: 0
    nsecs:         0
  frame_id: ''
goal_id: 
  stamp: 
    secs: 0
    nsecs:         0
  id: "1686130902134708426_/fetch_eus_interface_1686130898541295371_2753725_robotsound_jp_0"
goal: 
  sound_request: 
    sound: -3
    command: 1
    volume: 1.0
    arg: "\u3053\u3053\u304Cdock\u3067\u3059"
    arg2: "ja"

```

### DIFF
--- a/pr2eus/speak.l
+++ b/pr2eus/speak.l
@@ -43,6 +43,7 @@
           (when (eq (send ac :get-state) actionlib_msgs::GoalStatus::*active*)
             (send ac :cancel-goal)
             (send ac :wait-for-result :timeout 10))
+          (send goal :header :stamp (ros::time-now))
           (send goal :goal :sound_request msg)
           (setf (gethash topic-name *sound-play-clients*) ac)
           (send ac :send-goal goal)


k-okada:
@pazeshun I think ROS controller uses `goal.trajectory.header.stamp`, not `goal.header`

http://wiki.ros.org/pr2_controllers/Tutorials/Moving%20the%20arm%20using%20the%20Joint%20Trajectory%20Action#CA-11505f3470a9a076683bc46f9e917c9b05155c64_37

https://github.com/ros-controls/ros_controllers/blob/dce87855e1adf3a66271db8243d58561da9f2cdb/joint_trajectory_controller/include/joint_trajectory_controller/joint_trajectory_controller_impl.h#L610-L612

on pr2eus, we already set `start time` for both  `goal.trajectory.header.stamp` and `goal.header`


https://github.com/jsk-ros-pkg/jsk_pr2eus/blob/8fc634d0d92830c201f9ae922c5e673c0def0b19/pr2eus/robot-interface.l#L838-L853

tkmtnt7000:
I see. Thanks for your advice.

Now as long as I have checked, it seems that soundplay_node.py does not use timestamp of action goal topic, so I think this PR's change does not immediately have a side effect such as what you points out.
https://github.com/ros-drivers/audio_common/blob/3a748b223885a2f6035668c784e006e5426c84ef/sound_play/scripts/soundplay_node.py


FYI: 
What I originally wanted to do was to avoid that the timestamp would be 1970/1/1 if the timestamp was left at 0 when using the function to remember the robot's speech and recall it later. (as part of our recent efforts around database_talker)

PR #497, opened 2024-12-05
[pr2eus_moveit] align the center of primitive cylinder between eus and moveit

soonhyo:
The centor of primitvie cylinder is not matched between eus and moveit.

I added transformation code to align the center of the EUS model with the MoveIt model in the definition of the pose for the cylinder's solid primitive messages.

refer to following figures 

in eus 
Screenshot from 2024-12-05 15-27-03

in moveit (only cylinder models are not matched.)
Screenshot from 2024-12-05 15-27-59

PR
Screenshot from 2024-12-05 15-27-35

### DIFF
--- a/pr2eus_moveit/euslisp/collision-object-publisher.l
+++ b/pr2eus_moveit/euslisp/collision-object-publisher.l
@@ -78,7 +78,6 @@
                                   :dimensions (float-vector
                                                (/ (height-of-cylinder obj) 1000.0)
                                                (/ (radius-of-cylinder obj) 1000.0))))))
-         ;; Adjust the pose in the message to align EusLisp cylinder (origin at base) with MoveIt cylinder (origin at center).
          (send colobj :primitive_poses
                (append pose
                        (list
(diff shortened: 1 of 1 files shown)


pr2eus_moveit/euslisp/collision-object-publisher.l
```diff
@@ -81,7 +81,12 @@
          (send colobj :primitive_poses
                (append pose
                        (list
-                        (ros::coords->tf-pose (if relative-pose relative-pose (send obj :worldcoords))))))))
+                        (ros::coords->tf-pose (let ((coords (if relative-pose
+                                                             relative-pose
+                                                             (send obj :worldcoords))))
+                                               (let* ((hv (float-vector 0 0 (/ (height-of-cylinder obj) 2)))
```
pazeshun, line 87: - Why do you use `let` and `let*`? Is it impossible to use `let*` once?
- Please add a comment about the difference between euslisp cylinder's origin and ROS cylinder's origin.

pazeshun (changes requested):

pazeshun (approved):

PR #498, opened 2024-12-10
fix CI

soonhyo:
Same as https://github.com/jsk-ros-pkg/jsk_control/pull/790

Currently, CI fails with the same error as jsk_control:
https://github.com/jsk-ros-pkg/jsk_pr2eus/actions/runs/12192526626/job/34013294460

cc @pazeshun


pazeshun (approved):
CI passed.
@k-okada Could you review this?

pazeshun:
The same fix was merged via #500

PR #503, opened 2025-04-13
Add spin once option

soonhyo:
### Summary 
This PR adds a new keyword argument :spin-once to both :angle-vector and :send-ros-controller, allowing users to skip ROS groupname callbacks (ros::spin-once groupname). By setting :spin-once to nil, unnecessary robot state updates can be avoided, leading to significantly faster control loop frequencies.
Additionally, I modified :angle-vector in pr2_interface.l to support the :spin-once option. When :spin-once is set to nil, the calculation of diff-av for check-continuous-joint-move-over-180 uses the cached previous robot state in *ri* instead of calling ros::spin-once. (This has been tested in simulation but not on a real robot. I'm slightly concerned about cases where the initial state of the real robot significantly differs from the virtual robot's state.)

---
### Motivation
When ros::spin-once is executed for each groupname, it can introduce considerable latency, especially in a tight control loop. For instance, the loop frequency can drop to around 25Hz (as shown in the figure below) due to callback overhead.
In cases where :angle-vector is only used to send commands (without needing immediate state feedback), skipping these callbacks enables maintaining a loop rate of 500Hz or higher (as shown in the figure below). To support this, :spin-once was added as an optional parameter.
However, since fastest-tm calculation depends on up-to-date robot states, setting :spin-once to nil disables this. As a workaround, the time (tm) is explicitly set to 1000ms, which is assumed to be a reasonably safe default (though this value may need further validation).
Additionally, if tm is explicitly provided by the user while :spin-once is nil, it is accepted as-is — including values below 1000ms. While this may require more safety checks in the future, it's assumed the user can make reasonable decisions when prioritizing high-frequency loop control.

Figure_1
Figure_0

This is the test code I used when I made the plots.
```lisp
#!/usr/bin/env roseus
(require :pr2 "package://pr2eus/pr2-interface.l")
(ros::load-ros-package "geometry_msgs")
(ros::load-ros-package "std_msgs")

(unless (boundp '*pr2*) (pr2-init))

(defun log-benchmark-results (fn log-path log_prefix)
  (let ((results (bench2 (funcall fn))))
    (with-open-file (stream (format nil "log_pr2/~A_~A" log_prefix log-path) :direction :output :if-exists :append :if-does-not-exist :create)
        (format stream "~A~%" results))
    results))

(send *pr2* :reset-pose)
(send *ri* :angle-vector (send *pr2* :angle-vector))

(dotimes (i 100)
  (send *pr2* :rarm :move-end-pos #f(1 0 0) :world)
  (log-benchmark-results
     #'(lambda ()
         (send *ri* :angle-vector (send *pr2* :angle-vector) 1000 :default-controller 0 :spin-once t)
         )
     "log_angle_vector_raw.txt" "1"))

(send *pr2* :reset-pose)
(send *ri* :angle-vector (send *pr2* :angle-vector))

(dotimes (i 100)
  (send *pr2* :rarm :move-end-pos #f(1 0 0) :world)
  (log-benchmark-results
     #'(lambda ()
         (send *ri* :angle-vector (send *pr2* :angle-vector) 1000 :default-controller 0 :spin-once nil)
         )
     "log_angle_vector_raw.txt" "0"))

```
---
### Suggestions
- Further testing or evaluation may be needed to confirm whether 1000ms is a sufficiently safe default time when skipping state updates.
- The use of cached robot states in *ri* when :spin-once is nil should be validated, especially in scenarios where the real robot's initial state diverges significantly from the virtual state.
- In safety-critical applications, developers should be cautious when disabling callbacks and ensure that control logic does not rely on stale state information.

### DIFF
--- a/pr2eus/pr2-interface.l
+++ b/pr2eus/pr2-interface.l
@@ -408,16 +408,17 @@ Example: (send self :gripper :rarm :position) => 0.00"
        (incf i (send j :joint-dof)))
      add-new-trajectory-point))
   (:angle-vector
-   (av &optional (tm 3000) &rest args)
+   (av &optional (tm 3000) (ctype controller-type) (start-time 0) &key (spin-once t) &rest args)
    (let (diff-av)
-     ;; use reference-vector to get last commanded joint and use :angle-vector to toruncate the joint limit to eus lisp style
-     (setq diff-av (v- av (send robot :angle-vector (send self :state :reference-vector))))
+     ;; use reference-vector to get last commanded joint and use :angle-vector to truncate the joint limit to eus lisp style
+     (if spin-once
+         (setq diff-av (v- av (send robot :angle-vector (send self :state :reference-vector))))
+         (setq diff-av (v- av (send self :potentio-vector)))) ;; calculate joint difference based on the previous state
      ;; use shortest path for contiuous joint
-     ;;
      (when (send self :check-continuous-joint-move-over-180 diff-av)
        (return-from :angle-vector
          (send* self :angle-vector-sequence (list av) (list tm) args))) ;; when
-     (send-super* :angle-vector av tm args)
+     (send-super* :angle-vector av tm ctype start-time :spin-once spin-once args)
      ))
   (:angle-vector-sequence
    (avs &optional (tms (list 3000)) &rest args)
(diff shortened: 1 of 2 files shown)


knorth55:
I think in some code `ros::rate` is set with lower value.
You can change the value and check it.

https://github.com/jsk-ros-pkg/jsk_control/blob/42d1545e82c634390d1cf543eafa2288d607362a/eus_teleop/euslisp/lib/robot-teleop-interface.l#L956

